PR #18, opened 2024-09-16
Error free std.sol

d-xo:
Tweaks the compiler and std.sol so that executing `cabal run sol-core -- -f std/std.sol` no longer produces any errors.

Includes the following changes:

- tweaks the parse rules around identifiers to allow `_` in names
- support for hex literals in the lexer
- tweak the syntax / ast definitions for type synonyms to allow type parameters on the LHS
- adds the current lexeme to the parser error output
- updates `std.sol` to satisfy the current `sol-core` pipeline


d-xo:
I merged main and resolved conflicts but now I get type errors :(

```
Cannot match types:
a with n

 - in:function fst (p : Pair(a, b)) -> a {
   match (p) {
   | Pair(a, _) =>
      return a
   }
}
```

rodrigogribeiro (approved):
LGTM.

PR #19, opened 2024-09-16
concepts: sketch out has-field mechanism

d-xo:
adding a (currently non-parsing) sketch for generation of struct field access instances based on discussion in matrix.

### DIFF
--- a/concept-art/has-field.sol
+++ b/concept-art/has-field.sol
@@ -28,8 +28,8 @@ type sf3 = Unit
 
 // Field instances linking each field to it's position in the underlying tuple
 instance Pair(s, sf1):Field(Unit, uint) {}
-instance Pair(s, sf2):Field(Pair(uint, Unit), string) {}
-instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
+instance Pair(s, sf2):Field(uint, string) {}
+instance Pair(s, sf3):Field(Pair(uint, string), bool) {}
 
 
 // struct field member access desugars into calls to this class
--- a/concept-art/has-field.sol
+++ b/concept-art/has-field.sol
@@ -1,0 +1,52 @@
+data Unit = Unit
+data Pair(a, b) = Pair(a,b)
+
+type uint = word
+type string = word
+type bool = word
+
+data Memory(t) = Memory(Word)
+
+// this lets us link a given field in a struct to its position in it's
+// underlying generic representation as a tuple. 
+class self:Field(prevTypes, ty) {}
+
+// this struct should desugar into the following
+//struct S {
+//  f1 : uint;
+//  f2 : string;
+//  f3 : bool;
+//}
+
+// a type abstraction over tuples
+type s = Pair(uint, Pair(string, bool))
+
+// unique types identifying each field
+type sf1 = Unit
+type sf2 = Unit
+type sf3 = Unit
+
+// Field instances linking each field to it's position in the underlying tuple
+instance Pair(s, sf1):Field(Unit, uint) {}
+instance Pair(s, sf2):Field(Pair(uint, Unit), string) {}
+instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
+
+
+// struct field member access desugars into calls to this class
+class self:HasField(fieldType) {
+  function getField(x:sef) -> fieldType;
+}
+
+// we instantiate generic instances for references to types that implement Field
+instance (Pair(t, fieldName):Field(prevTypes, fieldType), fieldType:ValueType) => Pair(Memory(t), fieldName):HasField(fieldType) {
+  function getField(x : Pair(Memory(T), fieldName)) -> fieldType {
+    // TODO: define this function...
+    let x : Proxy(prevTypes) = Proxy;
+    let sz : Word = getMemorySize(x);
+    let ret : fieldType = ValueType.abs(0);
+    assembly {
+      ret := mload(add(rep(fst(x)), sz))
+    };
+    return ret;
+  }
+}
--- a/concept-art/has-field.sol
+++ b/concept-art/has-field.sol
@@ -34,11 +34,11 @@ instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
 
 // struct field member access desugars into calls to this class
 class self:HasField(fieldType) {
-  function getField(x:sef) -> fieldType;
+  function getField(x:self) -> fieldType;
 }
 
 // we instantiate generic instances for references to types that implement Field
-instance (Pair(t, fieldName):Field(prevTypes, fieldType), fieldType:ValueType) => Pair(Memory(t), fieldName):HasField(fieldType) {
+instance (Pair(t, fieldName):Field(prevTypes, fieldType), fieldType:ValueType) => Pair(Memory(t), fieldName):HasField(Memory(fieldType)) {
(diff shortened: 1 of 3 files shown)


concept-art/has-field.sol
```diff
@@ -0,0 +1,52 @@
...
+instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
+
+
+// struct field member access desugars into calls to this class
+class self:HasField(fieldType) {
+  function getField(x:sef) -> fieldType;
+}
+
+// we instantiate generic instances for references to types that implement Field
+instance (Pair(t, fieldName):Field(prevTypes, fieldType), fieldType:ValueType) => Pair(Memory(t), fieldName):HasField(fieldType) {
```
ekpyron: I think it should probably actually be

```suggestion
instance (Pair(t, fieldName):Field(prevTypes, fieldType), fieldType:ValueType) => Pair(Memory(t), fieldName):HasField(Memory(fieldType)) {
```

such that we can also assign *to* the thing in memory, not only read it (the implementation then just needs to bump the memory pointer of ``x.first`` by ``getMemorySize`` of ``prevTypes`` and we defer the actual load from memory.

concept-art/has-field.sol
```diff
@@ -0,0 +1,52 @@
...
+
+// Field instances linking each field to it's position in the underlying tuple
+instance Pair(s, sf1):Field(Unit, uint) {}
+instance Pair(s, sf2):Field(Pair(uint, Unit), string) {}
+instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
+
+
+// struct field member access desugars into calls to this class
+class self:HasField(fieldType) {
+  function getField(x:sef) -> fieldType;
```
ekpyron:
```suggestion
  function getField(x:self) -> fieldType;
```

concept-art/has-field.sol
```diff
@@ -0,0 +1,52 @@
...
+
+// unique types identifying each field
+type sf1 = Unit
+type sf2 = Unit
+type sf3 = Unit
+
+// Field instances linking each field to it's position in the underlying tuple
+instance Pair(s, sf1):Field(Unit, uint) {}
+instance Pair(s, sf2):Field(Pair(uint, Unit), string) {}
+instance Pair(s, sf3):Field(Pair(uint, Pair(string, Unit)), bool) {}
```
ekpyron: In the second two we don't really need the additional Unit:
```suggestion
instance Pair(s, sf2):Field(uint, string) {}
instance Pair(s, sf3):Field(Pair(uint, string), bool) {}
```

Also not sure if for the first one, we won't need a special
```
type FirstField = Unit;
instance Pair(s, sf1):Field(FirstField, uint) {}
```
Since I'm not sure if e.g. for ``Unit``s in storage we should actually assign a unique slot or the like (maybe not, but maybe better to have the option)

concept-art/has-field.sol
ekpyron: The signature of the ``getField`` function and its implementation would also need to change then, though :-).

concept-art/has-field.sol
```diff
@@ -0,0 +1,52 @@
...
+  function getField(x : Pair(Memory(T), fieldName)) -> fieldType {
+    // TODO: define this function...
+    let x : Proxy(prevTypes) = Proxy;
+    let sz : Word = getMemorySize(x);
+    let ret : fieldType = ValueType.abs(0);
+    assembly {
+      ret := mload(add(rep(fst(x)), sz))
+    };
+    return ret;
+  }
```
ekpyron, line 51:
```suggestion
  function getField(x : Pair(Memory(T), fieldName)) -> Memory(fieldType) {
    // TODO: define this function...
    let x : Proxy(prevTypes) = Proxy;
    let sz : Word = getMemorySize(x);
    let ret : Word = 0; // does this work or what am I supposed to initialize with here?
    assembly {
      ret := add(rep(fst(x)), sz)
    };
    return Memory(ret);
  }
```
something like this maybe (replacing https://github.com/ethlts/solcore/pull/19#discussion_r1761688002)

ekpyron:
Not sure if it's better or not - but I'd also be curious if something like this might work:

```solidity

//struct S {
//    x:uint256;
//    y:bool;
//    z:word;
//}
type S = Pair(uint256,Pair(bool, word));


// We might actually want to replace ``S.abs`` and ``S.rep`` by instances to such a class anyways.
class absType:Rep(repType) {
    function rep(x:absType) -> repType;
    function abs(x:repType) -> absType;
}
instance S:Rep(Pair(uint256,Pair(bool, word))) {
    function rep(x) { return S.rep(x); }
    function abs(x) { return S.abs(x); }
}


type Zero = Unit;
type Suc(n) = Unit;

class self:MemorySize {
    function size(_:Proxy(self)) -> word;
}
instance uint256:MemorySize {
    function size(_) -> word {
        return 32;
    }
}
instance bool:MemorySize {
    function size(_) -> word {
        return 32;
    }
}
instance word:MemorySize {
    function size(_) -> word {
        return 32;
    }
}
instance a:MemorySize => b:MemorySize => Pair(a,b):MemorySize {
    function size(_) -> word {
        let ap:Proxy(a) = Proxy;
        let asz = MemorySize.size(ap);
        let bp:Proxy(b) = Proxy;
        let bsz = MemorySize.size(bp);
        assembly { asz := add(asz, bsz) }
        return asz;
    }
}

class self:MemoryOffset {
    function get_offset(_:Proxy(self)) -> word;
}

type memoryOffset(tuple, index) = Unit;

instance memoryOffset(tuple, Zero):MemoryOffset {
    function get_offset(_) {
        return 0;
    }
}

instance a:MemorySize => memoryOffset(Pair(a,b), Suc(z)):MemoryOffset {
    function get_offset(_:Proxy(MemoryOffset(t,z))) {
        let sz = MemorySize.size(Proxy:Proxy(a));
        let offset = MemoryOffset.get_offset(_:Proxy(MemoryOffset(b, z)));
        assembly { offset := add(offset, sz) }
        return offset;
    }
}


type sf1 = Unit;
type sf2 = Unit;
type sf3 = Unit;

type Member(ty, field) = ty;

class self:Field(Index, Type) {}
instance Member(S, sf1):Field(Zero, uint256) {}
instance Member(S, sf2):Field(Suc(Zero), bool) {}
instance Member(S, sf3):Field(Suc(Suc(Zero)), word) {}

// Note: Assumes we have appropriate instances like: Memory(uint):Ref(uint)

instance ty:Rep(repType) => Member(ty,fieldName):Field(index, fieldType) => Member(Memory(ty), fieldName):Ref(fieldType)
{
    function load(x : Member(Memory(ty), fieldName)) -> fieldType {
        let base = Memory.rep(x.first);
        let p: Proxy(memoryOffset(repType, index)) = Proxy;
        let offset = MemoryOffset.get_offset(p);
        assembly {
            base := add(base, offset)
        }
        return Ref.load(Memory.abs(base):Memory(fieldType));
    }
    function store(x : Member(Memory(ty), fieldName), v:fieldType) {
        let p: Proxy(memoryOffset(repType, index));
        let base : Word = Memory.rep(x.first);
        let offset : Word = get_offset(p);
        assembly {
            base := add(base, offset)
        }
        Ref.store(Memory.abs(base):Memory(fieldType), v);
    }
}


// where
//   let x:S;
//   x.f1 = ...;
// becomes sugar for
//   let x:S;
//   Member.abs(x):Member(S,sf1) = ...;


```

rodrigogribeiro (approved):
LGTM

PR #22, opened 2024-09-19
Type-checking struct member example.

ekpyron:
This actually already type-checks as written.

However, I think our implementation of the Paterson Condition is incorrect and can be relaxed safely (we seem to count the sum over *all* context elements, whereas IIUC the condition is meant to be *per individual context element*, see https://github.com/ethlts/solcore/issues/23) - as well as the Coverage Condition can be relaxed safely, but lacking the relaxation of these conditions and the ability of disabling them per ``pragma``, for now I added an additional dead type argument to ``MemberAccess`` and artificially equipped it with a type that makes the crucial instance valid. However, given the right adjustments, this should keep working when removing this dead type argument.


rodrigogribeiro (approved):
LGTM.

PR #29, opened 2024-09-28
Changes to parsing and basic name resolution.

rodrigogribeiro:
Following Daniel's suggestion, the basic `word` type follows the lower case convention. Because of that, some changes were made on file `EmitCore.hs`.


mbenke:
Fails CI:

```
[31 of 37] Compiling Solcore.Frontend.Parser.SolcoreParser ( src/Solcore/Frontend/Parser/SolcoreParser.hs, /home/runner/work/solcore/solcore/dist-newstyle/build/x86_64-linux/ghc-9.8.2/sol-core-0.0.0.0/build/Solcore/Frontend/Parser/SolcoreParser.o, /home/runner/work/solcore/solcore/dist-newstyle/build/x86_64-linux/ghc-9.8.2/sol-core-0.0.0.0/build/Solcore/Frontend/Parser/SolcoreParser.dyn_o )

src/Solcore/Frontend/Parser/SolcoreParser.hs:20:34: error: [GHC-76037]
    Not in scope: type constructor or class ‘Name’
   |
20 |         | HappyAbsSyn4 (CompUnit Name)
   |                                  ^^^^
```

apparently `SolcoreParser.hs` is out of sync with `SolCoreParser.y` (do we need SolcoreParser.hs in the tree?)

mbenke:
also there seems to be a problem with `test/examples/cases/PairMatch1.solc`

mbenke:
After removing `SolcoreParser.hs` and merging `main`, the project builds on CI, but the `PairMatch1` test fails - see #30

rodrigogribeiro:
> also there seems to be a problem with `test/examples/cases/PairMatch1.solc`

Yes, I'm aware of that. I'm working on the fix which is to introduce skolemization of type variables in all annotated polymorphic types.

mbenke:
> > also there seems to be a problem with `test/examples/cases/PairMatch1.solc`
> 
> Yes, I'm aware of that. I'm working on the fix which is to introduce skolemization of type variables in all annotated polymorphic types.

I am not sure it's about skolemization - it looks like the typechecker is treating "word" as a type variable in `let x: word = p`

PR #34, opened 2024-10-09
Std typechecks

d-xo:
Tweaks the standard library so that it typechecks and runs it in ci. I had to comment out one function that gave me a type error I didn't understand, but we can address that in a separate pr / issue.

### DIFF
--- a/std/std.sol
+++ b/std/std.sol
@@ -34,17 +34,16 @@ function snd(p: Pair(a, b)) -> b {
     };
 }
 
-//
+// Proxy Type for Passing Types as Paramaters
 
 data Proxy(t) = Proxy;
 
 // References
 
-data Calldata(t) = Calladata(word);
-data Memory(t) = Memory(word);
-data Storage(t) = Storage(word);
-data Returndata(t) = ReturnData(word);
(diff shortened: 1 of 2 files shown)


std/std.sol
```diff
@@ -34,184 +34,193 @@ function snd(p: Pair(a, b)) -> b {
...
+    return mem;
 }
 
 instance (l:Encode, r:Encode) => Pair(l,r):Encode {
-    function shouldEncodeDynamic(x : Proxy(Pair(l,r))) -> Bool {
+    function shouldEncodeDynamic(x : Proxy(Pair(l,r))) -> bool {
         let l: Proxy(l) = Proxy;
         let r: Proxy(r) = Proxy;
-        return and(shouldEncodeDynamic(l), shouldEncodeDynamic(l));
+        return and(Encode.shouldEncodeDynamic(l), Encode.shouldEncodeDynamic(l));
```
aarlt, line 165: I just scrolled through the code.. isn't this a typo here? shouldn't this expression be `and(Encode.shouldEncodeDynamic(l), Encode.shouldEncodeDynamic(r));`?

PR #43, opened 2024-10-29
Rename variables in Yul assembly blocks

mbenke:
When translating to Yul, each variable/argument in Sol code is mapped to one or several Yul variables, consequently also word-size variables get renamed. This PR has variables in Yul assembly blocks renamed in the same way so that word-sized variables from Sol code can be accessed in assembly blocks. This also allows for unsafe casts - we should discuss if this is a good thing.

An alternative approach would be to prevent renaming of simple variables when translating to Yul, but it seems more complicated and fragile.


mbenke:
At the call we decided not to go this route, but rather keep, and make accessible in assembly blocks, only the names of variables of type word.

I will create another PR implementing this.

PR #47, opened 2024-11-28
Implementation of closure conversion.

rodrigogribeiro:
This PR provides the implementation of the closure conversion algorithm. Additionally, this PR provides bug fixes and syntax sugar for tuples.


runsol.sh
mbenke, line 1: Why is it empty?

src/Solcore/Frontend/Parser/SolcoreParser.y
```diff
@@ -308,13 +312,14 @@ Pattern : Name PatternList                         {Pat $1 $2}
         | '_'                                      {PWildcard}
         | Literal                                  {PLit $1}
         | '(' Pattern ')'                          {$2}
+        | PatternList                              {Pat (Name "pair") $1}
```
mbenke, line 315: Is it always a pair? what about `(x, y, z)` ?

test/examples/cases/Compose.solc
mbenke, line 1: Why was main removed?

test/examples/cases/NegPair.solc
mbenke, line 1: Whe remove main, bnot, fromB?

src/Solcore/Frontend/Parser/SolcoreParser.y
rodrigogribeiro: I use the name `pair` to denote tuples, not only pairs on the syntax.

test/examples/cases/Compose.solc
rodrigogribeiro: During the debugging, I removed the `main`from this test case. I added the following code:

```
function main () {
   let f = compose(id,id);
  return f(42);
}
```

test/examples/cases/NegPair.solc
rodrigogribeiro: I removed during debugging. I've restored this file to its old version.

runsol.sh
rodrigogribeiro: Fixed.

PR #48, opened 2024-11-29
Definition of a pragma for boolean type.

rodrigogribeiro:
Initial implementation for a pragma for setting up the boolean type. Current syntax definition:

```
data bool = true | false;

pragma bool-type-decl bool, true, false;
```

### DIFF
--- a/src/Solcore/Frontend/Pretty/SolcorePretty.hs
+++ b/src/Solcore/Frontend/Pretty/SolcorePretty.hs
@@ -1,18 +1,20 @@
 {-# LANGUAGE InstanceSigs #-}
-module Solcore.Frontend.Pretty.SolcorePretty(module Common.Pretty, pretty) where
 
-import qualified Data.List.NonEmpty as N 
+module Solcore.Frontend.Pretty.SolcorePretty (module Common.Pretty, pretty) where
+
 import Data.List
+import Data.List.NonEmpty qualified as N
 
 import Prelude hiding ((<>))
 
-import Solcore.Frontend.Syntax.Contract
-import Solcore.Frontend.Syntax.Name 
 import Solcore.Frontend.Pretty.Name
-import Solcore.Frontend.Syntax.Stmt 
+import Solcore.Frontend.Syntax.Contract
+import Solcore.Frontend.Syntax.Name
+import Solcore.Frontend.Syntax.Pragma
+import Solcore.Frontend.Syntax.Stmt
 import Solcore.Frontend.Syntax.Ty
 import Solcore.Frontend.TypeInference.Id
-import Solcore.Frontend.TypeInference.TcSubst 
+import Solcore.Frontend.TypeInference.TcSubst
 
 import Common.Pretty
 import Language.Yul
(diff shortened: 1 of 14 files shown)


mbenke:
What's the intended meaning? To translate declared true/false to Yul true/false/ Anything else? 
Perhaps also if/while when we have them? Maybe we should think how they should work.

rodrigogribeiro:
> What's the intended meaning? To translate declared true/false to Yul true/false/ Anything else? Perhaps also if/while when we have them? Maybe we should think how they should work.

That's my main question: What should I do with this information? Should I add if / while statements to the surface syntax?

src/Solcore/Frontend/Syntax/Pragma.hs
```diff
@@ -0,0 +1,32 @@
...
+  | NoBoundVariableCondition
+  deriving (Eq, Ord, Show, Data, Typeable)
+
+-- status for type inference pragmas
+
+data PragmaStatus
+  = Enabled
+  | DisableAll
+  | DisableFor (NonEmpty Name)
+  deriving (Eq, Ord, Show, Data, Typeable)
```
mbenke, line 22: I don't know what `PragmaStatus` is, please add comment in code

mbenke (changes requested):
It's very hard to tell what changed when actual changes are mixed with big reformatting. Could you please split this PR in two (reformatting, actual pragma stuff)?

mbenke:
Something is missign in prettyprinter:

```
$ cabal run -- sol-core -v -f test/examples/pragmas/booltypesyntax.solc
> AST after name resolution
sol-core: src/Solcore/Frontend/Pretty/SolcorePretty.hs:(56,3)-(58,46): Non-exhaustive patterns in function ppr
```

src/Solcore/Frontend/Pretty/SolcorePretty.hs
```diff
@@ -21,318 +23,321 @@ import Language.Yul
...
+  ppr (TPragmaDecl p) = ppr p
 
-instance Pretty Pragma where 
-  ppr (Pragma _ Enabled) = empty 
-  ppr (Pragma ty st) 
-    = hsep [text "pragma", ppr ty, ppr st, semi] 
+instance Pretty Pragma where
+  ppr (InferencePragma _ Enabled) = empty
+  ppr (InferencePragma ty st) =
+    hsep [text "pragma", ppr ty, ppr st, semi]
```
mbenke, line 58:
```suggestion
instance Pretty Pragma where
  ppr (InferencePragma _ Enabled) = empty
  ppr (InferencePragma ty st) =
    hsep [text "pragma", ppr ty, ppr st, semi]
  ppr (BoolTypePragma (ty N.:| cons)) =
    text "pragma bool-type-decl" <+> commaSep (map ppr (ty:cons))
    
```

d-xo:
> > What's the intended meaning? To translate declared true/false to Yul true/false/ Anything else? Perhaps also if/while when we have them? Maybe we should think how they should work.
> 
> That's my main question: What should I do with this information? Should I add if / while statements to the surface syntax?

IIRC we need this for at least two things:

1. short circuiting for boolean ops (this will need compiler magic since we're in a strict eval context)
2. surface level `if`/`for`/`while`

d-xo:
> It's very hard to tell what changed when actual changes are mixed with big reformatting. Could you please split this PR in two (reformatting, actual pragma stuff)?

Agree, this is impossible to review as is. @rodrigogribeiro how much work would it be to tidy this up? We could also just close it if it's significant effort, since this pr is not needed for the stuff we're focusing on right now.

rodrigogribeiro:
> > It's very hard to tell what changed when actual changes are mixed with big reformatting. Could you please split this PR in two (reformatting, actual pragma stuff)?
> 
> Agree, this is impossible to review as is. @rodrigogribeiro how much work would it be to tidy this up? We could also just close it if it's significant effort, since this pr is not needed for the stuff we're focusing on right now.

Most of these changes have been made in other commits. The implementation of Boolean type was left aside due other more pressing fixes on type inference. I'll close this PR, since it does not make sense currently.

PR #50, opened 2024-12-03
Generate Yul objects

mbenke:
First stab at generating Yul objects (no dispatch/deploy, everything in init code as we discussed)

# Usage
## Compiling

    $ cabal exec -- sol-core -f test/examples/spec/022add.solc

(writes to output.core)

### Using HEVM

    $ cabal run yule -- output.core -o Input.yul
    $ solc --strict-assembly --bin --optimize Input.yul | tail -1 > Output.hex
    $ hevm exec --code $(cat Output.hex) | awk -f parse_hevm_output.awk
    Return: 42
    hex: 0x000000000000000000000000000000000000000000000000000000000000002a

The file `testsol.sh` defines bash functions `hevmcore` and `hevmsol` that perform these steps from core and solc source respectively

### Using Foundry

    $ cabal run yule -- output.core -w -o Output.sol
      found main
      writing output to Output.sol
    $ forge script Output.sol
    Script ran successfully.
    Gas used: 24910

    == Logs ==
    RESULT -->  42

### DIFF
--- a/src/Language/Yul.hs
+++ b/src/Language/Yul.hs
@@ -18,13 +18,6 @@ newtype Yul = Yul { yulStmts :: [YulStmt] }
 newtype YulCode = YulCode YulBlock
 
 instance {-# OVERLAPPABLE #-} Pretty a => Show a where show = render . ppr
-{-
-instance Show Yul where show = render . ppr
-instance Show YulStmt where show = render . ppr
-instance Show YulExp where show = render . ppr
-instance Show YLiteral where show = render . ppr
-instance Show YulData where show = render . ppr
--}
 
 instance Semigroup Yul where
   Yul a <> Yul b = Yul (a <> b)
@@ -163,31 +156,3 @@ instance Pretty HexOrString where
   ppr (DHex s) = text "hex" <> doubleQuotes (text s)
   ppr (DString s) = doubleQuotes (text s)
 
-
--- sample code
-fnUsrAdd :: YulStmt
-fnUsrAdd = YFun "usr$add" ["x", "y"] (YReturns ["result"])
-        [ YAssign ["result"] (YCall "add" [YIdent "x", YIdent "y"])]
-
-fnUsrMain :: YulStmt
-fnUsrMain = YFun "usr$main" [] (YReturns ["result"])
-        [ YAssign1 "result" (YCall "usr$add" [yulInt 40, yulInt 2])]
-
-sampleCode :: YulCode
-sampleCode = YulCode
-            [ fnUsrAdd
-            , fnUsrMain
-            , YulAlloc "z"
-            , YAssign1 "z" (YCall "usr$main" [])
-            , YExp $ YCall "mstore" [yulInt 0, YIdent "z"]
-            , YExp $ YCall "return" [yulInt 0, yulInt 32]
-            ]
-
-sampleObject :: YulObject
-sampleObject = YulObject "Add" sampleCode []
-
-
-sampleNestedObject :: YulObject
-sampleNestedObject = YulObject "Nested" sampleCode
-                        [  InnerObject sampleObject
-                        , InnerData $ YulData "Table1" (DHex "4123")]
--- a/src/Language/Yul.hs
+++ b/src/Language/Yul.hs
@@ -166,34 +166,6 @@ instance Pretty HexOrString where
 -- commaSepList :: Pretty a => [a] -> Doc
 -- commaSepList = hsep . punctuate comma . map ppr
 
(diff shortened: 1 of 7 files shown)


src/Language/Yul.hs
```diff
@@ -124,33 +156,40 @@ instance Pretty YLiteral where
...
+            ]
+
+sampleObject :: YulObject
+sampleObject = YulObject "Add" sampleCode []
+
+
+sampleNestedObject :: YulObject
+sampleNestedObject = YulObject "Nested" sampleCode
+                        [  InnerObject sampleObject
+                        , InnerData $ YulData "Table1" (DHex "4123")]
```
d-xo: Is this needed?

src/Language/Yul.hs
```diff
@@ -1,18 +1,30 @@
...
+newtype YulCode = YulCode YulBlock
+
+instance {-# OVERLAPPABLE #-} Pretty a => Show a where show = render . ppr
+{-
 instance Show Yul where show = render . ppr
 instance Show YulStmt where show = render . ppr
 instance Show YulExp where show = render . ppr
 instance Show YLiteral where show = render . ppr
+instance Show YulData where show = render . ppr
+-}
```
d-xo: Do we need this?

src/Language/Yul.hs
```diff
@@ -124,33 +156,40 @@ instance Pretty YLiteral where
...
 
+instance Pretty YulData where
+  ppr (YulData name val) = hsep [text "data", doubleQuotes $ text name , ppr val]
+
+instance Pretty HexOrString where
+  ppr (DHex s) = text "hex" <> doubleQuotes (text s)
+  ppr (DString s) = doubleQuotes (text s)
+
 -- commaSepList :: Pretty a => [a] -> Doc
 -- commaSepList = hsep . punctuate comma . map ppr
```
d-xo:
```suggestion
```

yule/Main.hs
```diff
@@ -32,8 +35,56 @@ main = do
...
+      [ call "mstore" [yulInt 0, YIdent "_mainresult"]
+      , call "return" [yulInt 0, yulInt 32]
+      ]
+    call f args = YExp (YCall f args)
+
+{- | wrap a Yul chunk in a Solidity function with the given name
+   assumes result is in a variable named "_result"
+-}
+
+wrapInSol name yul = wrapInContract name "wrapper()" wrapper
```
d-xo, line 60: Do we need this since we have the yul object now?

d-xo (approved):
just a few questions / nits, but otherwise lgtm.

yule/Main.hs
mbenke: > Do we need this since we have the yul object now?

Yul objects should probably be optional until we find a good way of running them (the hevm way described above is not really convenient - it seems that `forge script` is  easier for now. But I am open to suggestions.

d-xo: whats the issue with the hevm approach? we can write a custom harness or use another another tool if more features are required.

src/Language/Yul.hs
mbenke: Removed

yule/Main.hs
mbenke: > whats the issue with the hevm approach? we can write a custom harness or use another another tool if more features are required.

Once someone writes the custom harness, we can easily switch.

d-xo: my question is: what is needed that is currently missing?

src/Language/Yul.hs
mbenke: Removed

mbenke:
> I think it should be a goal to move over to it and get rid of our foundry dependency.
> 
> What do you think still needs to be done before we can make yof the default?

We need a driver/script to run complete pipeline and print the results. Let's discuss it at the next meeting, I have put it in the agenda.

testsol.sh
d-xo, line 1: Is the intended usage that I would source this file?

testsol.sh
```diff
@@ -0,0 +1,56 @@
...
+    forge script Output.sol | grep RESULT
+}
+
+function testspec() {
+    file=$1
+    echo $file
+    shift
+    cabal exec sol-core -- -f $file --debug-spec --dump-spec $*
+#    cabal run yule -- output.core -O && \
+#    forge script Output.sol
```
d-xo, line 23: Is this supposed to be commented?

mbenke: Actually `testspec` is just to debug the cases where specialisation fails. Hence they are usually not executed.

PR #54, opened 2024-12-13
Adding unit type back / fixes on pretty printing.

rodrigogribeiro:
* Adding the unit type back, which solve some ambiguities which are present when treating this type as 0-tuple.
* Fixes function pretty printing, which was wrongly dealing with type variables and constraints.

### DIFF
--- a/src/Solcore/Frontend/Pretty/SolcorePretty.hs
+++ b/src/Solcore/Frontend/Pretty/SolcorePretty.hs
@@ -142,6 +142,11 @@ instance Pretty a => Pretty (Signature a) where
     = text "function" <+> 
       ppr n           <+>
       pprParams ps    <+> 
+      pprRetTy ty
+  ppr (Signature _ [] n ps ty) 
+    = text "function" <+> 
+      ppr n           <+>
+      pprParams ps    <+> 
       pprRetTy ty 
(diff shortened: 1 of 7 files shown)


src/Solcore/Frontend/Pretty/SolcorePretty.hs
```diff
@@ -316,8 +321,12 @@ instance Pretty Ty where
     = ppr t1 <+> (text "->") <+> ppr t2
   ppr (TyCon n ts)
     | isTuple n = parens $ commaSep (map ppr ts)
+    | isUnit n = text "()"
```
mbenke, line 324: How should unit be spelled in source files? "unit" or "()"?

mbenke (changes requested):
It would be better with a (simple) test

mbenke (approved):

mbenke:
What about parsing?

also please see that test/examples/cases/Ref.solc  is updated

rodrigogribeiro:
Fixed in the last commit.

src/Solcore/Frontend/Pretty/SolcorePretty.hs
rodrigogribeiro: Fixed. Notation is `()`.

PR #61, opened 2025-01-10
DRAFT: PoC of Ref.load/store specialisation without intrinsics, but using "pending statements" instead

mbenke:


mbenke:
Abandoning this approach in favour of stkLoad/stkStore builtins (#60)

PR #62, opened 2025-01-14
Refactoring erasure to a different module

rodrigogribeiro:
This is the first of a series of PR's to fix the closure conversion algorithm.

PR #63, opened 2025-01-14
Refactoring unique type generator

rodrigogribeiro:
This moves the unique type logic for a dedicated module. This should be merged after #62.

### DIFF
--- a/src/Solcore/Desugarer/UniqueTypeGen.hs
+++ b/src/Solcore/Desugarer/UniqueTypeGen.hs
@@ -24,7 +24,7 @@ instance UniqueTypeGen a => UniqueTypeGen (Maybe a) where
 
 instance UniqueTypeGen (CompUnit Name) where 
   uniqueTyGen (CompUnit _ ds) 
-    = mapM_ uniqueTyGen ds 
+    = uniqueTyGen ds 
 
 instance UniqueTypeGen (TopDecl Name) where
   uniqueTyGen (TContr c) = uniqueTyGen c 
--- a/src/Solcore/Desugarer/UniqueTypeGen.hs
+++ b/src/Solcore/Desugarer/UniqueTypeGen.hs
@@ -1,0 +1,99 @@
+module Solcore.Desugarer.UniqueTypeGen where
+
+import Control.Monad.State 
+import qualified Data.Map as Map
+import Solcore.Frontend.Syntax
+import Solcore.Frontend.Pretty.SolcorePretty
+
+uniqueTypeGen :: CompUnit Name -> IO (CompUnit Name, UniqueTyMap)
+uniqueTypeGen c@(CompUnit imps ds)
+  = do 
+      env <- runUniqueM (uniqueTyGen c)
+      let ds' = (TDataDef <$> Map.elems (uniqueMap env)) ++ ds
+      pure (CompUnit imps ds', uniqueMap env)
+
+class UniqueTypeGen a where 
+  uniqueTyGen :: a -> UniqueM ()
+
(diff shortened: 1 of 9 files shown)


src/Solcore/Desugarer/UniqueTypeGen.hs
```diff
@@ -0,0 +1,99 @@
...
+instance UniqueTypeGen a => UniqueTypeGen [a] where
+  uniqueTyGen = mapM_ uniqueTyGen
+
+instance UniqueTypeGen a => UniqueTypeGen (Maybe a) where
+  uniqueTyGen Nothing = pure ()
+  uniqueTyGen (Just x) = uniqueTyGen x
+
+instance UniqueTypeGen (CompUnit Name) where 
+  uniqueTyGen (CompUnit _ ds) 
+    = mapM_ uniqueTyGen ds
```
mbenke:
```suggestion
    = uniqueTyGen ds 
```

src/Solcore/Frontend/TypeInference/Erase.hs
```diff
@@ -0,0 +1,79 @@
...
+  erase :: a -> EraseRes a
+
+instance Erase a => Erase [a] where
+  type EraseRes [a] = [EraseRes a]
+  erase = map erase
+
+instance Erase a => Erase (Maybe a) where
+  type EraseRes (Maybe a) = Maybe (EraseRes a)
+  erase Nothing = Nothing
+  erase (Just x) = Just (erase x)
```
mbenke:
```suggestion
  erase  = fmap erase
```

src/Solcore/Frontend/TypeInference/Erase.hs
```diff
@@ -0,0 +1,79 @@
...
+  type EraseRes (Pat Id) = Pat Name
+
+  erase (PVar n)
+    = PVar (idName n)
+  erase (PCon n ps)
+    = PCon (idName n) (erase ps)
+  erase PWildcard
+    = PWildcard
+  erase (PLit l)
+    = PLit l
```
mbenke, line 76: There should be a less verbose way to do it (using syb maybe?) but we can let it be.

test/Main.hs
```diff
@@ -79,7 +79,7 @@ cases
               , expectFail $ runTestForFile "DupFun.solc" caseFolder
               , runTestForFile "DuplicateFun.solc" caseFolder
               , runTestForFile "EitherModule.solc" caseFolder
-              , expectFail $ runTestForFile "Id.solc" caseFolder
+              , runTestForFile "Id.solc" caseFolder
```
mbenke, line 82: It's still not completetely working:
```
PANIC: spec(let f : t_lambda_impl01(word, n) = id() ;): free type vars in t_lambda_impl01(word, n)
```

This is due to the fact that main gets desugared to
```
   function main () -> word {
      let f : t_lambda_impl01(word, n) = id() ;
      return invokable.invoke(f, 0);
```

while I think it probably should be `let f : t_lambda_impl01(word, n)`

mbenke (approved):

src/Solcore/Frontend/TypeInference/Erase.hs
rodrigogribeiro: I thought to use syb, but I was trying to do it quick since there are another major changes. Maybe, later I could change this to use syb and avoid so much boilerplate code.

test/Main.hs
mbenke: Ah, I see it's probably fixed in #65. Well done :)

PR #75, opened 2025-01-28
Fixes on context reduction for the references example.

rodrigogribeiro:
* Introduces instantiation for instances during context reduction.
* Adding test case for references


src/Solcore/Frontend/TypeInference/TcMonad.hs
```diff
@@ -405,6 +415,7 @@ toHnf depth pred@(InCls n _ _)
                   ++"\nKnown instances:\n"++ (unlines $ map pretty is))
         Just (preds, subst', instd) -> do
             extSubst subst'
+            x <- getSubst
```
mbenke, line 418: I believe this is unnecessary

```suggestion
```

mbenke (approved):
Well done!

Congratulations

PR #76, opened 2025-02-05
no-desugar-calls option; make options available in the typechecker monad

mbenke:
Adds the option `--no-desugar-calls` that disables indirect call desugaring.
It also makes program options available in the typechecking monad (TcM) via `getTcOpts`
Options aremoved to a separate `Solcore.Pipeline.Options` module.


rodrigogribeiro (approved):
Great! Thank you!

PR #77, opened 2025-02-05
Some tools for instance resolution debugging

mbenke:
Main additions:
- `tcmError` - when verbose, dump logs before throwing an error 
- `tcmMgu` - wrap `mgu` call in `tcmError`
- some more logging for instance resolution


rodrigogribeiro (approved):
Great! Thanks, it will be very helpful!

Issue #36, opened 2024-10-09
Instance Method Can be Called on a Fully Unconstrained Variable 


d-xo:
The following passes typechecking but I think it should fail (since `t` is not constrained to be an instance of `ValueTy` anywhere:

```
data memory(t) = memory(word);

class t:ValueTy {
    function rep(x:t) -> word;
}

instance memory(t) : ValueTy {
    function rep(x: memory(t)) -> word {
        match x {
        | memory(w) => return w;
       };
    }
}

class ref:Ref(deref) {
    function store(loc: ref, value: deref) -> Unit;
}

instance memory(t) : Ref(t) {
    function store(loc: memory(t), value: t) -> Unit {
        // We don't have a `ValueTy` bound on `t` anywhere, so this should raise a type error...
        let vw = ValueTy.rep(value);
    }
}
```

mahmudsudo:
hi , can i take on this ?

mbenke:
Solved by #85

PR #85, opened 2025-02-08
Introduces entailment and fixes instance generation.

rodrigogribeiro:
* This PR implements a constraint entailment test, adds super class support and fixes some issues on `invokable` instance generation. 

* Added some tests cases for super class support and entailment.

This fixes issue #36

### DIFF
--- a/src/Solcore/Frontend/Syntax/Ty.hs
+++ b/src/Solcore/Frontend/Syntax/Ty.hs
@@ -44,13 +44,25 @@ splitTy t = ([], t)
 funtype :: [Ty] -> Ty -> Ty
 funtype ts t = foldr (:->) t ts
 
-alphaEq :: Ty -> Ty -> Bool 
-alphaEq (TyVar _) (TyVar _) 
-  = True 
-alphaEq (TyCon n ts) (TyCon n' ts')
-  = n == n' && (and (zipWith alphaEq ts ts'))
-alphaEq _ _ 
-  = False
+class AlphaEq a where 
+  alphaEq :: a -> a -> Bool
+
+instance AlphaEq a => AlphaEq [a] where 
+  alphaEq ts ts' = and $ zipWith alphaEq ts ts'
+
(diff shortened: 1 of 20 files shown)


src/Solcore/Frontend/Syntax/Ty.hs
```diff
@@ -44,13 +44,26 @@ splitTy t = ([], t)
...
-  = False
+class AlphaEq a where 
+  alphaEq :: a -> a -> Bool
+
+instance AlphaEq a => AlphaEq [a] where 
+  alphaEq ts ts' = and $ zipWith alphaEq ts ts'
+
+instance AlphaEq Ty where 
+  alphaEq (TyVar _) (TyVar _) 
+    = True
```
mbenke, line 55: Should this really always be True ? Is `a->b =α= a->a` ?

src/Solcore/Frontend/TypeInference/InvokeGen.hs
```diff
@@ -90,7 +79,9 @@ createInstance udt fd sch
         scall = Return (Call Nothing fname ssargs)
         bdy = Match [discr] [([foldr1 ppair (spvs : sargs)], [scall])] 
         ifd = FunDef isig [bdy]
-      pure (Instance qs invokableName [argTy, retTy] selfTy [ifd]) 
+        instd = Instance qs invokableName [argTy, retTy] selfTy [ifd]
+      info [">> Generated invoke instance:\n", pretty instd]
```
mbenke:
```suggestion
      info [">> Generated invokable instance:\n", pretty instd]
```

mbenke (approved):

src/Solcore/Frontend/Syntax/Ty.hs
rodrigogribeiro: You are right. This is **VERY** wrong. Thanks for pointing this.

PR #98, opened 2025-04-09
Measure time of some phases

mbenke:
Daniel was complaining about compilation times so I have added some diagnostic timing (behind the `--timing` option) to find the culprits, but this may come handy later.


rodrigogribeiro (approved):
LGTM.

PR #100, opened 2025-04-09
Fixes the unit instance bug. Changes syntax on example test cases.

rodrigogribeiro:
This fixes the bug in instance definitions using the unit type, `()`.


mbenke (approved):
Looks good!

PR #101, opened 2025-04-09
Adding an option for disabling match compilation.

rodrigogribeiro:
Adding an option to disable pattern match compilation. This is useful to reduce to output produced by the prototype.


mbenke (approved):
LGTM

PR #102, opened 2025-04-09
Fixes on context reduction.

rodrigogribeiro:
This PR fixes a problem in context reduction. Now, the following code is correctly rejected by the type inference.

```

class abs:Typedef(rep) {
    function abs(v:rep) -> abs;
    function rep(v:abs) -> rep;
}

instance memory(a):Typedef(word) {
    function abs(ptr:word) -> memory(a) {
        return memory(ptr);
    }
    function rep(v:memory(a)) -> word {
        match v {
            | memory(ptr) => return ptr;
        };
    }
}

pragma no-patterson-condition Test;
pragma no-bounded-variable-condition Test;
class self:Test {
    function test(x:self) -> word;
}

instance word:Test {
    function test(x:word) -> word {
        return x;
    }
}

data test(a) = test(memory(a));

instance test(a):Typedef(memory(a)) {
    function rep(x:test(a)) -> memory(a) {
        match x {
            | test(m) => return m;
        };
    }
    function abs(m:memory(a)) -> test(a) {
        return test(m);
    }
}

forall test(abs):Typedef(rep), rep:Test . instance test(abs):Test {
    function test(x:test(abs)) -> word {
        return Test.test(Typedef.rep(x));
    }
}

contract C {
    function main() {
        let x:test(word) = test(memory(42));
        let ptr:word = Test.test(x);
    }
}
```

### DIFF
--- a/src/Solcore/Frontend/Pretty/SolcorePretty.hs
+++ b/src/Solcore/Frontend/Pretty/SolcorePretty.hs
@@ -298,8 +298,10 @@ instance Pretty Pred where
 instance Pretty Scheme where
   ppr (Forall vs ty) = ppr' (Forall vs ty) 
     where 
+      ppr' (Forall [] ([] :=> t)) = ppr t
       ppr' (Forall [] (ctx :=> t))
-        = pprContext ctx <+> ppr t
+        = text "forall"       <+>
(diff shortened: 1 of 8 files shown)


mbenke (approved):
Well done!

PR #103, opened 2025-04-10
Update pretty-printing of type schemes to current syntax

mbenke:
Print type schemes with nontrivial context and no variables as

```
forall context . t
```
e.g.

```
forall memory(structType) : Typedef (o10) . word
```

rather than

```
memory(structType) : Typedef (o10) . word
```

which is more consistent with the current syntax (e.g. `forall constraints . instance ...`)


rodrigogribeiro (approved):
LGTM. Thanks!

PR #109, opened 2025-04-15
Fixes on class/instance syntax. Fixes on test case syntax.

rodrigogribeiro:
This PR fixes the syntax of class and instance definitions and remove the semi colon from assembly and match blocks.
Note that semicolon still needed on lambda blocks appearing at the RHS of an assingment like:

```
let id = lam(x) {return x ; };
```


mbenke (approved):
I have fixed syntax of some missed files in the spec folder. Otherwise, looks good.

PR #110, opened 2025-04-17
Fix on the pretty-print and parser to use new syntax. Small fixes on type checker.

rodrigogribeiro:
* Updates the syntax of examples.
* Fix the pretty print for the new syntax.
* Adds context reduction for local definitions.
* Fix a typo on context reduction.


mbenke (commented):
The content does not match the description - it includes changes to the typechecker, parser and also some temporary LaTeX files

src/Solcore/Frontend/TypeInference/TcReduce.hs
```diff
@@ -26,7 +26,7 @@ splitContext ps fs
       let (ds, rs) = partition (all (`elem` fs) . fv) ps'
       info [">> Defered constraints:", pretty ds]
       info [">> Retained constraints:", pretty rs]
-      pure (ds, ps)
+      pure (ds, rs)
```
mbenke: What is this change? Is it the typo you have mentioned in the call?

mbenke (approved):
Needs conflict resolution. Otherwise looks good.

src/Solcore/Frontend/TypeInference/TcReduce.hs
rodrigogribeiro: Yes! This is the typo... basically it was returning the original constraint set instead of the retained ones.

PR #111, opened 2025-04-29
Implementation of default instances.

rodrigogribeiro:


mbenke (approved):
Needed for #113; we will do more extensive testing in 113

PR #112, opened 2025-05-08
enable use of sstore in assembly

mbenke:


d-xo:
changes to foundry lib are superseded by #120. @mbenke ok if we remove those changes from this pr?

mbenke:
@d-xo [b186413](https://github.com/argotorg/solcore/pull/112/commits/b186413232cb95b3607f9a879a3ed5508ace5a79) might be still worth keeping (or we need a better way of handling the modifier to allow sload/sstore when needed)

d-xo:
yes, sorry just meant backing out the lib changes and keeping the sstore stuff

d-xo (approved):

PR #122, opened 2025-05-31
Add forge-std as dependency

r0qs:
This is an alternative solution to https://github.com/argotorg/solcore/pull/120 adding forge-std as submodule dependency.

It requires to run `forge update` to download and update the submodule.


rodrigogribeiro (approved):
LGTM.

PR #124, opened 2025-05-31
Add forge to nix environment

r0qs:
I'm not a Nix expert, but I attempted to fix https://github.com/argotorg/solcore/issues/117. Let me know if this is not the right/best approach ;)


rodrigogribeiro (approved):
Thanks!

PR #130, opened 2025-06-16
Specialise: rename  type variables, report ambiguous types

mbenke:


src/Solcore/Desugarer/Specialise.hs
```diff
@@ -104,6 +113,13 @@ flex t = t
 -- make all type variables flexible in a syntactic construct
 flexAll :: Data a => a -> a
 flexAll = everywhere (mkT flex)
+-}
+
+-- | A signature forall tvs . t is considered ambiguous if `tvs \\ FTV(t) /= mempty`
+-- this is should be the same as `FTV(body) \\ FTV(t) /= {}`
+-- returns list of ambiguous variables
+ambiguousVarsInSig :: HasTV a => Signature a -> [Tyvar]
```
d-xo, line 120: Could this be a syntactic check enforced by requiring explicit forall everywhere as we've discussed a few times previously?

mbenke: But are we going to require full type signatures for every function? If not, then syntactic check is not enough, since this checks the *output* of the typechecker.

mbenke: Currently fails tests due to Issue #136

d-xo:
what is the status here? do we still want this?

PR #132, opened 2025-06-17
Fix nix build

d-xo:
`nix build` was previously failing since `cabal` was not available in the nix sandbox at test time.

This PR reworks the tests so that they invoke the solver and solcore pipelines directly, avoiding the calls to cabal. I reworked the solcore pipeline so that argument parsing and file writing happens in a wrapper, and then restructured it to return an `Either` and flattened a lot of the control flow.

CI now runs `nix build`, meaning we can exactly replicate the full pipeline locally.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,22 @@
 {
   "nodes": {
+    "ethereum-tests": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1698656397,
+        "narHash": "sha256-1zhBaJ3X5kQ94qv9Yz12/3d83w3jNP6OtzjH+ykK8sg=",
+        "owner": "ethereum",
+        "repo": "tests",
+        "rev": "428f218d7d6f4a52544e12684afbfe6e2882ffbf",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ethereum",
+        "ref": "v13",
+        "repo": "tests",
+        "type": "github"
+      }
+    },
     "flake-utils": {
       "inputs": {
         "systems": "systems"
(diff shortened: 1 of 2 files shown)


src/Solcore/Pipeline/SolcorePipeline.hs
```diff
@@ -29,93 +35,100 @@ pipeline :: IO ()
...
   startTime <- Time.getCurrentTime
   opts <- argumentsParser
+  result <- compile opts
+  case result of
+    Left err -> do
+      putStrLn err
+      exitWith (ExitFailure 1)
+    Right contracts -> do
+      forM_ (zip [1..] contracts) $ \(i, c) -> do
+        let filename = "output" <> show i <> ".core"
```
d-xo, line 44: we would previously overwrite the same output.core file in the multiple contract case. this change does mean that the default output file is `output1.core`.

mbenke: Well done! One question - the CI on Github gives this warning:

```
warning: download buffer is full; consider increasing the 'download-buffer-size' setting
```

do you know what to do with it? Does it come from Nix? Or Github?

d-xo: It's from nix. Happens because the runner is slow and it has fast internet. I think it's pretty cosmetic and can be ignored.

https://github.com/NixOS/nix/pull/11171

mbenke (approved):
Good work. Can you please just resolve conflicts and perhaps add hevm dependency now that you want it to be the default way of running programs?

PR #133, opened 2025-06-24
Core/Yul: fix handling of parentheses

mbenke:


src/Language/Core/Parser.hs
```diff
@@ -93,6 +93,7 @@ pPrimaryExpr = choice
     , pTuple
     , try (ECall <$> identifier <*> parens (commaSep coreExpr))
     , EVar <$> (identifier  <* notFollowedBy (symbol "("))
+    , parens coreExpr
```
d-xo, line 96: lets just accept valid yul

src/Language/Yul/Parser.hs
```diff
@@ -55,6 +55,7 @@ yulExpression = choice
     [ YLit <$> yulLiteral
     , try (YCall <$> pName<*> parens (commaSep yulExpression))
     , YIdent <$> pName
+    , parens yulExpression
```
clonker: This parses invalid Yul

PR #138, opened 2025-06-25
InvokeGen: Handle closure-only argument lists

d-xo:
We can get into a state where a function takes only closure arguments if it captures from the env but takes no other arguments. This would result in the generation of incorrect patterns when trying to match against the `(self, args)` pair in the invoke method.

The fix modifies createInstance to additionaly insert a `unit` into the args list in the case where all args are closure types, ensuring generation of correct patterns in the closure only case.


rodrigogribeiro (approved):
Great! Thanks!

PR #141, opened 2025-06-26
Support Cyclic Function Definitions

d-xo:
This PR includes two changes that fix compilation of mutually recursive functions:

1. The strongly connected component analysis was not wrapping connected binding groups in the `TMutualDef` and `CMutualDecl` constructors, meaning that they were not being passed into `tcBindGroup` and were being checked individually.
2. `tcBindGroup` was not adding the other functions in the bind group to the environment when typechecking.

For now we scoped the fix to work for strongly connected definition groups composed only of functions. Expanding support to include classes and instances will need a bit more work (since tcBindGroup currently only supports typechecking function definitions).

### DIFF
--- a/src/Solcore/Frontend/TypeInference/SccAnalysis.hs
+++ b/src/Solcore/Frontend/TypeInference/SccAnalysis.hs
@@ -79,8 +79,6 @@ analysis fn ds
 
 -- building the dependency graph
 
--- building the dependency graph
-
 mkGraph :: (Ord a, Names a, Decl a) => [a] -> AdjacencyMap a
 mkGraph ds = stars $ mkEdges (mkNameEnv ds) ds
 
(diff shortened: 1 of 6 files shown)


src/Solcore/Frontend/TypeInference/SccAnalysis.hs
```diff
@@ -38,28 +38,44 @@ sccAnalysis' :: CompUnit Name -> SCC (CompUnit Name)
...
+        then singleton (CMutualDecl xs)
+        else xs
+
+      isFunDef (CFunDecl _) = True
+      isFunDef _ = False
 sccContract d = pure d
 
-analysis :: (Ord a, Names a, Decl a) => [a] -> SCC [a]
-analysis ds
+analysis :: (Ord a, Names a, Decl a, Show a) => ([a] -> [a]) -> [a] -> SCC [a]
```
d-xo: we couldn't find a clean way to make this generic, so ended up implementing `group` twice with concrete types and passing it in. sugggestions welcome.

mbenke:
The specialiser did not handle mutual recursion (`CmutualDecl/TMutualDef` nodes). Should be fixed now.

PR #149, opened 2025-07-06
Fix the logging of matching compilation. 

rodrigogribeiro:
* Fixes the wrong condition on logging the result of pattern matching compilation.
* Removes an unnecessary type matching on tcFunDef.


mbenke (approved):
LGTM

PR #157, opened 2025-07-25
TcMonad: faster freshName using Data.Set

mbenke:
This small change speeds up typechecking a lot.

Profiling has shown that 75% of the time is used by GC and the main culprit is `freshName`.
So I have rewritten it using Data.Set rather than lists and voila :)


rodrigogribeiro (approved):
Great job! Thanks!

rodrigogribeiro (approved):
LGTM.

PR #164, opened 2025-08-11
ci: verbose nix build output

d-xo:
Prints the full build /test log in ci


clonker (approved):
Very nice! Sorry for missing earlier that it does indeed run the tests. In the future we could perhaps have the tests run as an individual CI step, that way one doesn't have to dig through the logs as much. Optional in any case.

PR #167, opened 2025-08-18
Implementing if statement

rodrigogribeiro:
* Adds if to high level syntax
* Desugar if's to match constructs
* Desugar the `bool` type into `sum unit unit`
** Desugar `true` to `inl unit`.
** Desugar `false` to `inr unit`.

### DIFF
--- a/test/Cases.hs
+++ b/test/Cases.hs
@@ -47,9 +47,8 @@ spec =
     , runTestForFile "939badfood.solc" specFolder
     , runTestForFile "SimpleField.solc" specFolder
     , runTestForFile "121counter.solc" specFolder
-    -- Failing due to missing assign constraint
     , runTestExpectingFailureWith stdOpt { optNoDesugarCalls = True } "126nanoerc20.solc" specFolder
-  --, runTestForFile "127microerc20.solc" specFolder
+    , runTestExpectingFailure "127microerc20.solc" specFolder
     ]
  where
   specFolder = "./test/examples/spec"
@@ -158,6 +157,7 @@ cases =
     , runTestExpectingFailure "nano-desugared.solc" caseFolder
     , runTestExpectingFailure "uintdesugared.solc" caseFolder
     , runTestForFile "word-match.solc" caseFolder
+    , runTestForFile "if-examples.solc" caseFolder
     ]
  where
   caseFolder = "./test/examples/cases"


mbenke:
I can't see any examples/tests. Could you add some?

src/Solcore/Desugarer/IfDesugarer.hs
```diff
@@ -0,0 +1,64 @@
...
+
+desugarStmt :: Stmt Id -> Stmt Id
+desugarStmt (Match es eqns)
+  = Match es (map (\ (ps, bdy) -> (ps, map desugarStmt bdy)) eqns)
+desugarStmt (If e bdy1 bdy2)
+  = Match [e] [eqntrue, eqnfalse]
+    where
+      eqntrue = ([PCon (Id inlName (inlTy unit unit)) [PCon (Id "()" unit) []]], bdy1)
+      eqnfalse = ([PCon (Id inrName (inrTy unit unit)) [PCon (Id "()" unit) []]], bdy2)
+desugarStmt v = v
```
d-xo, line 31: We don't need to recurse into any other statements? Maybe `Let` and `Return`?

src/Solcore/Desugarer/IfDesugarer.hs
```diff
@@ -0,0 +1,64 @@
...
+  | otherwise = Con c (map desugarBoolCons es)
+desugarBoolCons (FieldAccess me v)
+  = FieldAccess (desugarBoolCons <$> me) v
+desugarBoolCons (Call me v es)
+  = Call (desugarBoolCons <$> me) v (map desugarBoolCons es)
+desugarBoolCons (Lam ps bdy ty)
+  = Lam ps (everywhere (mkT desugarBoolCons) bdy) ty
+desugarBoolCons (TyExp e t)
+  = TyExp (desugarBoolCons e) t
+desugarBoolCons v = v
```
d-xo: Small nitpick and feel free to ignore, but I slightly prefer the following since it's more robust in case we decide to expand `Exp` in the future with additional cases (since it won't fall through to the default without warning).

```suggestion
desugarBoolCons (Var a) = Var a
desugarBoolCons (Lit l) = Lit l
```

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -464,6 +478,11 @@ tcFunDef incl vs' qs d@(FunDef sig@(Signature vs ps n args rt) bd)
       -- building the function type scheme
       free <- getEnvMetaVars
       (ds, rs) <- splitContext ps1' (ps1 ++ qs1) free
+      unless (null ds) $ do
+            (tcmError $ unlines [ "Cannot entail:"
+                                , pretty ds
+                                , "from"
+                                , pretty sig]) `wrapError` d
```
d-xo: Why is this needed?

src/Solcore/Pipeline/SolcorePipeline.hs
```diff
@@ -94,11 +96,20 @@ compile opts = runExceptT $ do
     putStrLn "> Elaborated tree:"
     putStrLn $ pretty typed
 
+  -- If / boolean desugaring
+
+  typed' <- liftIO $ if noIfDesugar then pure typed
```
d-xo: maybe `desugared` instead of `typed'`?

src/Solcore/Desugarer/IfDesugarer.hs
rodrigogribeiro: No, because the only place that has a recursion on the statement tree is in `if` and `match` blocks: 

```
data Stmt a
  = (Exp a) := (Exp a)                  -- assignment
  | Let a (Maybe Ty) (Maybe (Exp a))    -- local variable
  | StmtExp (Exp a)                     -- expression level statements
  | Return (Exp a)                      -- return statements
  | Match [Exp a] (Equations a)         -- pattern matching
  | Asm YulBlock                        -- Yul block
  | If (Exp a) (Body a) (Body a)        -- If statement
  deriving (Eq, Ord, Show, Data, Typeable)
```

There is the possibility of lambda expressions which can have if's inside. But since the main recursion is done through the function `everywhere` from the scrap your boilerplate package, it will apply the function `desugarStmt` in every statement in the syntax tree. I'll add a test covering this possibility.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: This is removed in #171 .

src/Solcore/Desugarer/IfDesugarer.hs
rodrigogribeiro: I added a new function to the test case which uses an if inside a lambda expression.

d-xo (approved):

PR #174, opened 2025-08-28
Small fixes to dispatch.sol

mbenke:


d-xo:
I think we can close this in favour of #175?

PR #176, opened 2025-09-03
First version of string literals handling

mbenke:
Quick and dirty string literal handling - literals longer than 32 chars are truncated

### DIFF
--- a/test/Cases.hs
+++ b/test/Cases.hs
@@ -42,6 +42,7 @@ spec =
     , runTestForFile "06comp.solc" specFolder
     , runTestForFile "09not.solc" specFolder
     , runTestForFile "092revert.solc" specFolder
+    , runTestForFile "093longString.solc" specFolder
     , runTestForFile "10negBool.solc" specFolder
     , runTestForFile "11negPair.solc" specFolder
     , runTestForFile "903badassign.solc" specFolder


src/Language/Core/Parser.hs
```diff
@@ -124,7 +124,7 @@ coreStmt = choice
     , SFunction <$> (pKeyword "function" *> identifier) <*> (parens (commaSep coreArg)) <*> (symbol "->" *> coreType)
                 <*> (symbol "{" *> many coreStmt <* symbol "}")
     , SAssembly <$> (pKeyword "assembly" *> yulBlock)
-    , SRevert <$> (pKeyword "revert" *> stringLiteral)
```
d-xo, line 127: why?

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -262,7 +262,7 @@ emitExp (Var x) = do
         Just e -> pure (e, [])
         Nothing -> pure (Core.EVar (unwrapId x), [])
 -- special handling of revert
-emitExp (Call _ (Id "revert" _) [Lit(StrLit s)]) = pure(Core.EUnit, [Core.SRevert s])
```
d-xo, line 265: Is this supposed to be commented?

src/Solcore/Frontend/Syntax/ElabTree.hs
```diff
@@ -728,7 +736,36 @@ instance Elab S.Literal where
...
-  elab (S.StrLit s) = pure (StrLit s)
+  elab (S.StrLit s)
+     | length s <= 32 = elabStr s
+     | otherwise = do
+         let s' = take 32 s
+         writes ["Warning: long string literal truncated to:\n  ", show s']
+         elabStr s'
+     where
+       -- elabStr = pure . IntLit . encodeString
+       elabStr = fmap IntLit . debugString
```
d-xo: can we put the debug output behind the `--verbose` flag or something?

test/Cases.hs
```diff
@@ -41,6 +41,7 @@ spec =
     , runTestForFile "048rgb2.solc" specFolder
     , runTestForFile "06comp.solc" specFolder
     , runTestForFile "09not.solc" specFolder
+    , runTestForFile "092revert.solc" specFolder
```
d-xo, line 44: we should also add the longString test here.

src/Language/Core/Parser.hs
mbenke: Don't know if we need revert builtin in core, but I renamed it to `__revert__`, because otherwise you cannot define a function called `revert`

src/Solcore/Desugarer/EmitCore.hs
mbenke: Yes, I have disabled special handling of revert.
Another option might be to keep special  handling of revert and transform string literals on the Core <-> Yul stage.

But the current typechecker chokes on type literals, so I decided to do it in the preliminar desugaring stage, at least for now.

src/Solcore/Frontend/Syntax/ElabTree.hs
mbenke: Not easily, but I have disabled this debug.

test/Cases.hs
mbenke: done

mbenke:
For comnparison, I will try another approach, keeping string literals throughout the frontend and desugaring them in the backend.

Issue #179, opened 2025-09-04
Add support for desugared bool in EmitCore


mbenke:
```
$ cat test/examples/boolbug.solc
contract Bug {
  function main() {
    return false;
  }
}

$ esolc test/examples/boolbug.solc 
test/examples/boolbug.solc
Emitting core for contract Bug

ERROR: translateTCon: unknown type sum
sum
in: function signature function main () -> sum((), ())
in: function main

sol-core: Emit core failed
CallStack (from HasCallStack):
  error, called at src/Solcore/Desugarer/EmitCore.hs:41:3 in sol-core-0.0.0.0-inplace:Solcore.Desugarer.EmitCore
  errorsEM, called at src/Solcore/Desugarer/EmitCore.hs:181:20 in sol-core-0.0.0.0-inplace:Solcore.Desugarer.EmitCore
```

PR #183, opened 2025-09-05
EmitCore: handle desugared bools (solves #179)

mbenke:


src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -1,14 +1,12 @@
 module Solcore.Desugarer.EmitCore(emitCore) where
-import Prelude hiding(catch)
+import Prelude hiding(catch, product)
 import Language.Core qualified as Core
 import Data.Map qualified as Map
 import Common.Monad
-import Control.Monad(forM, when, unless)
-import Control.Monad.IO.Class
+import Control.Monad(when, unless)
+-- import Control.Monad.IO.Class
```
d-xo:
```suggestion
```

src/Language/Core/Parser.hs
```diff
@@ -70,12 +70,12 @@ pKeyword w = try $ lexeme (string w <* notFollowedBy identChar)
 
 pPrimaryType :: Parser Type
 pPrimaryType = choice
-    [ TWord <$ pKeyword "word"
+    [ try $ TNamed <$> identifier <*> braces coreType
```
d-xo, line 73: why? does this allow overriding the built in `word` or `bool` types?

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -51,7 +49,7 @@ data EcState = EcState
 initEcState :: Bool -> TcEnv -> EcState
 initEcState debugp env = EcState
    { ecSubst = emptyVSubst
-   , ecDT = Map.empty
+   , ecDT = Map.fromList builtinDataInfo
```
d-xo, line 51: just for my understanding, what is an `ecDT`?

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -103,10 +112,11 @@ emitTopDecl _ = pure []
 addData :: DataTy -> EM ()
 addData dt = modify (\s -> s { ecDT = Map.insert (dataName dt) dt (ecDT s) })
 
+{-
```
d-xo, line 114:
```suggestion
```

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -69,6 +67,17 @@ type DataTable = Map.Map Name DataTy
...
+sumDataTy = DataTy
+  { dataName = "sum"
+  , dataParams = [TVar "a", TVar "b"]
+  , dataConstrs = [ Constr "inl" [tyvar "a"]
+                  , Constr "inr" [tyvar "b"]
+                  ]
+  } where
+    tyvar = TyVar . TVar
+
+builtinDataInfo = [ ("sum", sumDataTy) ]
```
d-xo, line 79: should there also be a `pair` here?

src/Solcore/Desugarer/EmitCore.hs
mbenke: `ec`is the "emit core" prefix, and `DT` stands for `DataTable`

src/Solcore/Desugarer/EmitCore.hs
mbenke: Popssibly; until now `pair` was handled specially, because it used to be variadic (i.e. could denoted any n-tuple). IIRC now tuples are desugared to nested pair somewhere earlier.

Do you think it's worthwhile to try to remove the special cases?

src/Language/Core/Parser.hs
mbenke: `sum` was a keyword, which led to user-defineed (or in this case compiler-defined) type named `sum` to fail to parse. We probably need a better syntax, but this fix works for now. 

I dont'want to devote more time to this now, since
eventually, `yule` should probably be integrated with the rest of the compiler and we will not need to pass the IL (aka Core) via text files so the concrete syntax will be mostly irrelevant.

src/Solcore/Desugarer/EmitCore.hs
d-xo: > Do you think it's worthwhile to try to remove the special cases?

Definitely not in this pr. I don't have enough context to have a strong opinion on that in general. Maybe lets open an issue so we don't forget?

PR #184, opened 2025-09-05
Desugar true/false in patterns (fixes #182)

### LINKED ISSUE #182: true and false not desugared in matches
For the program 

```
function not(x:bool) -> bool {
  match x {
    | true => return false;
    | false => return true;
  }
}
```

we do not desugar true and false in match:

```
$ esolc test/examples/matchboolbug.solc -s --dump-ds
test/examples/matchboolbug.solc
> Match compilation result:
forall self args ret . class  self : invokable (args, ret) {
   forall self args . function invoke (self<self> : self, args<args> : args) -> ret;
}
function not (x<sum((), ())> : sum((), ())) -> sum((), ()) {
   match (x<sum((), ())>) {
   | false<sum((), ())> =>
      return inl<() -> sum((), ())>(()<()>);
   | true<sum((), ())> =>
      return inr<() -> sum((), ())>(()<()>);
   }
}
```

which leads to problems when emitting core (reverts are emitted because there are no matches for inl/inr)

mbenke:


rodrigogribeiro (approved):
Thanks!

PR #186, opened 2025-09-06
Adding string type info into initial type environment

rodrigogribeiro:


src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -472,7 +472,7 @@ tcFunDef incl vs' qs d@(FunDef sig@(Signature vs ps n args rt) bd)
           qs1 = everywhere (mkT (insts @Ty env)) qs
       -- getting argument / return types in annotations
       (args', lctx, ts') <- tcArgs args1
-      rt1' <- maybe freshTyVar pure rt1
+      rt1' <- maybe freshTyVar kindCheck rt1
```
d-xo, line 475: why? is this related to the string fix?

rodrigogribeiro: In fact, this should be done before, since we need to check if the return type is valid, i.e., all mentioned types are in fact defined and type applications should respect the correct type constructor arity.

PR #192, opened 2025-09-09
Fixes on bound variable condition test

rodrigogribeiro:
This PR provides the following fixes:

- Fix the bound variable condition test for class and instance definitions.
- Adds checks for kinds in instance definitions.
- Check if a class is defined in instance definitions.
- Adds a pragma in std.solc
- Adds some test cases proposed by @dxo.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -747,7 +747,7 @@ checkInstance idef@(Instance d vs ctx n ts t funs)
       unless patterson (checkMeasure ctx ipred `wrapError` idef)
       -- checking bound variable condition
       bound <- askBoundVariableCondition n
-      unless bound (checkBoundVariable ctx (fv (t : ts)) `wrapError` idef)
+      unless bound (checkBoundVariable ctx (bv (t : ts)) `wrapError` idef)
       -- checking instance methods
       mapM_ (checkMethod ipred) funs `wrapError` idef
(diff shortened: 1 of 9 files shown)


src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -741,9 +745,18 @@ checkCompleteInstDef n ns
...
+checkConstraints = mapM_ checkConstraint
+
 checkInstance :: Instance Name -> TcM ()
 checkInstance idef@(Instance d vs ctx n ts t funs)
   = do
+      -- kind check all types in instance head
+      mapM_ kindCheck (t : ts) `wrapError` idef
+      -- check if the class is defined
+      _ <- askClassInfo n `wrapError` idef
+      -- check if the context has types valid with classes
```
```suggestion
      -- check if all the types and classes in the context are valid
```

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -780,7 +793,7 @@ isTyVar _ = False
 
 checkBoundVariable :: [Pred] -> [Tyvar] -> TcM ()
 checkBoundVariable ps vs
-  = unless (all (\ v -> v `elem` vs) (fv ps)) $ do
+  = unless (all (\ v -> v `elem` vs) (bv ps)) $ do
```
```suggestion
  = unless (all (`elem` vs) (bv ps)) $ do
```

mbenke (approved):
Well done. 
Could we also rename the pragma itself to `no-bound-variable-condition` (it is `NoBoundVariableCondition` in the sources) while we are at it? Or maybe in a separate PR?

PR #195, opened 2025-09-16
Fixes on instance typing

### LINKED ISSUE #189: overlapping instances accepted by compiler
The following snippet typechecks but should not (note the argument type to the methods for the `()` instance).

```
data uint256 = uint256(word);
forall self . class self:ABIAttribs {`
    function headSize(ty:Proxy(self)) -> word;
    function isStatic(ty:Proxy(self)) -> bool;
}

instance ():ABIAttribs {
    function headSize(ty : Proxy(uint256)) -> word { return 0; }
    function isStatic(ty : Proxy(uint256)) -> bool { return true; }
}
instance uint256:ABIAttribs {
    function headSize(ty : Proxy(uint256)) -> word { return 32; }
    function isStatic(ty : Proxy(uint256)) -> bool { return true; }
}
```

rodrigogribeiro:
This PR fixes #189, but it also found some problems on dispatch on std.solc.


test/Cases.hs
```diff
@@ -10,7 +10,8 @@ std :: TestTree
 std =
   testGroup
     "Standard library"
-    [ runTestForFile "std.solc" stdFolder
+    -- this should work, but neet fix on the dispatch
+    [ runTestExpectingFailure "std.solc" stdFolder
```
d-xo: the dispatch logic in std.solc is deleted in https://github.com/argotorg/solcore/pull/201 btw

rodrigogribeiro:
> what was wrong with the previous implementation?

The problem is that type inference was not checking if the annotated type in the instance is compatible with its expected type. 
Since this was not checking, you could have something like:

```
forall a . class a: Foo {
    function foo (x : a) -> word;
}

instance word : Foo {
    function foo (x : word) -> word {
         return x;
    }
}

instance bool : Foo {
   function foo (x : word) -> word {
       return 0;
  }
}
```

The `bool` instance should be rejected since it is expected to have a function of type `bool -> word` and not `word -> word`, as it is currently annotated.

test/Cases.hs
```diff
@@ -157,10 +158,10 @@ cases =
...
     , runTestForFile "word-match.solc" caseFolder
     , runTestForFile "if-examples.solc" caseFolder
-    , runTestForFile "import-std.solc" caseFolder
-    , runTestExpectingFailure "bound-minimal.solc" caseFolder
-    , runTestExpectingFailure "bound-only-test.solc" caseFolder
-    , runTestForFile "bound-with-pragma.solc" caseFolder
+    , runTestExpectingFailure "import-std.solc" caseFolder
+    , runTestExpectingFailure "withdraw.solc" caseFolder
+    , runTestForFile "bal.solc" caseFolder
+    , runTestForFile "if-examples.solc" caseFolder
```
d-xo: why stop running the bound-X tests?

test/Cases.hs
```diff
@@ -157,10 +158,10 @@ cases =
     , runTestForFile "uintdesugared.solc" caseFolder
     , runTestForFile "word-match.solc" caseFolder
     , runTestForFile "if-examples.solc" caseFolder
-    , runTestForFile "import-std.solc" caseFolder
```
d-xo, line 161: why did this start failing?

test/Cases.hs
d-xo: can this be changed to pass again now the dispatch logic is removed?

test/examples/cases/instance-fun-typing.solc
d-xo, line 1: Ia this test being run in ci? I don't see it in Cases.hs

std/std.solc
```diff
@@ -693,8 +693,8 @@ forall reader baseType reader baseType_decoded .baseType : ABIAttribs, reader:Wo
         let prx : Proxy(baseType_decoded);
         let result : memory(DynArray(baseType_decoded)) = allocateDynamicArray(prx, length);
         let offset = 0;
-        let prx : Proxy(baseType);
-        let elementHeadSize : word = ABIAttribs.headSize(prx);
+        // let prx : Proxy(baseType);
+        // let elementHeadSize : word = ABIAttribs.headSize(prx);
```
d-xo: why?

test/examples/cases/mydispatch.solc
d-xo, line 1: why do we need this file? can't we just merge the real std/dispatch.solc and test that? what is different here?

test/examples/cases/mystd.solc
d-xo, line 1: what is different here vs std/std.sol. why do we need both?

test/Cases.hs
```diff
@@ -167,8 +167,8 @@ cases =
     , runTestForFile "tuva.solc" caseFolder
     , runTestForFile "yul-return.solc" caseFolder
     , runTestForFile "pragma_merge_base.solc" caseFolder
-    , runTestForFile "pragma_merge_import.solc" caseFolder
-    , runTestForFile "pragma_merge_verify.solc" caseFolder
+    , runTestExpectingFailure "pragma_merge_import.solc" caseFolder
+    , runTestExpectingFailure "pragma_merge_verify.solc" caseFolder
```
d-xo: why?

rodrigogribeiro: The file `pragma_merge_import` does have overlapping instances:

```
forall X .
X : TestClassP2, X : TestClassP3, X : TestClassP4 => instance ImportType2 : TestClassP1 {
}
forall U .
U : TestClassP1, U : TestClassP2, U : TestClassP3 => instance U : TestClassP1 {
}
```
The same happens with `pragma_merge_verify.solc`.

test/examples/cases/mystd.solc
rodrigogribeiro: I've added the files `mystd.solc` and `mydispatch.solc` to test the verification of the issue of accepting instances with the same head.

test/examples/cases/mydispatch.solc
rodrigogribeiro: I've added the files mystd.solc and mydispatch.solc to test the verification of the issue of accepting instances with the same head.

std/std.solc
rodrigogribeiro: I commented these because they do not influence on the `result` of the function. Removing the comments, it will work just fine. I'll restore this to its previous state.

test/examples/cases/mydispatch.solc
d-xo: can we replace them with some minimal test cases that cover the issue? these are going to be annoying to maintain going forward.

test/Cases.hs
d-xo: Ah, can we fix the tests so they no longer contain overlapping instances instead of just marking them as failing please?

test/examples/cases/mydispatch.solc
rodrigogribeiro: Remove `mystd.solc` and `mydispatch.solc` and added a simple example.

test/Cases.hs
rodrigogribeiro: Fixed.

test/examples/cases/mystd.solc
rodrigogribeiro: I removed both files and added a simple test case for overlapping.

test/examples/cases/pragma_merge_verify.solc
```diff
@@ -7,7 +7,7 @@ import pragma_merge_base;
 data VerifyType(x) = VerifyType;
 
 // Would fail without imported pragma no-patterson-condition TestClassP1
-forall A . A:TestClassP1, A:TestClassP2, A:TestClassP3 => instance A:TestClassP1 {}
+forall A . A:TestClassP1, A:TestClassP2, A:TestClassP3 => instance VerifyType(A):TestClassP1 {}
```
mbenke: This is OK wrt Paterson condition, so pointless as a test, also makes the comment above misleading.

test/examples/cases/pragma_merge_base.solc
```diff
@@ -26,7 +26,7 @@ data TestType2 = TestType2;
 // --- Patterson Condition ---
 
 // Fails Patterson: Context measure (3 predicates) > conclusion measure (TestType1(U):TestClassP1)
-forall U . U:TestClassP1, U:TestClassP2, U:TestClassP3 => instance U:TestClassP1 {}
+forall U . U:TestClassP1, U:TestClassP2, U:TestClassP3 => instance (U,word):TestClassP1 {}
```
mbenke: This is OK wrt Paterson condition, so pointless as a test, also makes the comment above misleading.

mbenke:
```suggestion
// Fails Patterson: context constraint not smaller then head
forall U . (U,word):TestClassP1 => instance U:TestClassP1 {}
```

test/examples/cases/pragma_merge_verify.solc
mbenke:
```suggestion
// Would fail without imported pragma no-patterson-condition TestClassP3
forall A . word:TestClassP3 => instance A:TestClassP3 {}
```

..and add `TestClassP3` to the pragma (using TestClassP1 leads to overlapping)

d-xo:
> > Can we also add a test for #189
> 
> I added a test case for that: `overlapping-heads.solc`

That seems to just test actual overlapping instances, and not cases where the instance sig is incorrect.

PR #197, opened 2025-09-16
Parser: handle yul return

d-xo:
return is used as both a surface level keyword and a yul level function. This modifies the surface level parser to correctly handle `return` in yul blocks.


d-xo:
test failures are unrelated, fixed in https://github.com/argotorg/solcore/pull/205.

PR #199, opened 2025-09-16
runsol.sh: use hevm, allow specifying calldata

d-xo:
This modifies the runsol.sh script to:

- use yul/hevm instead of solidity/forge (the inline yul variant in solidity does not support the full range of builtins we need).
- adds cli flags to allow specifying the callvalue/calldata passed to the contract at runtime


mbenke (changes requested):
Could you please add some usage notes to the README?

PR #200, opened 2025-09-16
TcContract: Allow definition of multiple pragmas

d-xo:
This is needed when importing a file that contains pragma definitions into another one. Pragmas are merged additively, with a catch all pragma defintion taking precendence and forcing the check to be disabled for all instances


rodrigogribeiro (approved):
LGTM.

PR #201, opened 2025-09-16
std: add instances for contract dispatch desugaring

d-xo:
This adds the core datatype definitions and class / instance definitions that allow us to desugar a surface level `contract` definition into a single entrypoint that dispatches to the contracts methods based on it's function selector hash.

An example desugaring can be seen at the end of the file. At a high level each surface level `contract` object will have a `main` generated that calls `RunContract.exec` on an instance of the `Contract` datatype reflecting the structure of the given contract. The defined typeclass instances will handle the dispatch generation and abi encoding/decoding.

Still to do:

- receive()
- move callvalue check to the top level if all methods are not payable
- `public` / `external`

Pending a decision on our implementation of compile time eval it also relies on the desugaring stage to precompute the selector hash and insert that directly into the bytecode.

You can test the desugaring logic as follows. This will call the `add2` method with `2` and `3` abi encoded, and then display the result:

```
$ ./runsol.sh std/dispatch.solc --calldata "add2(uint256,uint256)" 2 3
Processing: std/dispatch.solc
Compiling to core...
Emitting core for contract C
Writing to output1.core
Generating Yul...
found main
writing output to dispatch.yul
Compiling to bytecode...
Hex output: dispatch.hex
Executing...
"Return: 0x0000000000000000000000000000000000000000000000000000000000000005"
Decoded: 5
```

### DIFF
--- a/std/dispatch.solc
+++ b/std/dispatch.solc
@@ -21,17 +21,85 @@ data Fallback(payability, args, rets, fn) = Fallback(payability, args, rets, fn)
 
 // --- Method Selectors ---
 
-// For each method in a contract the compiler generates a unique type and
-// produces a `Selector` instance for that type that returns the selector hash
-forall nm . class nm:Selector {
-  function hash(prx: Proxy(nm)) -> word;
+// The string representation of abi types used during selector computation
+forall ty . class ty:ABIString {
+    function append(head : word, tail : word, prx : Proxy(ty)) -> word;
 }
 
-// Method has a Selector if its name has a Selector
-forall name payability args rets fn . name:Selector => instance Method(name,payability,args,rets,fn):Selector {
-  function hash(prx: Proxy(Method(name,payability,args,rets,fn))) -> word {
-    return Selector.hash(Proxy : Proxy(name));
-  }
+instance uint256:ABIString {
+    function append(head : word, tail : word, prx : Proxy(uint256)) -> word {
+        let size : word = 7;
+        assembly {
+            mstore(head, add(mload(head), size))
+            mstore(tail, 0x75696e7432353600000000000000000000000000000000000000000000000000)
+        }
+        return Add.add(tail, size);
+    }
+}
+
+instance ():ABIString {
+    function append(head : word, tail : word, prx : Proxy(())) -> word {
+        return(tail);
+    }
+}
+
+function append_left_bracket(head : word, tail : word) -> word {
+    let size = 1;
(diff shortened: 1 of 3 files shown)


std/dispatch.solc
```diff
@@ -0,0 +1,293 @@
+import std;
+
+pragma no-patterson-condition RunDispatch, ExecMethod;
+
+// --- Preliminaries ---
+
+data Bool = True | False;
```
d-xo: Using the builtin `bool` type gave me a bunch of specializer errors I didn't understand. I'll try to find a minimal reproducer and submit a bug.

std/dispatch.solc
```diff
@@ -0,0 +1,293 @@
...
+            | True => ExecMethod.exec(method_m);
+            | False => return ();
+          }
+        }
+    }
+  }
+}
+
+// Recursive instance
+forall n m . n:ExecMethod, n:Selector, m:RunDispatch => instance (n,m):RunDispatch {
```
d-xo, line 216: Should this be treated as overlapping with the instance above?

mbenke (changes requested):
For the record
- fails if a contract has more than one method
- problems with argumentless methods (e.g. `function one() -> uint256  { return uint256(1); }


It also would be good to split the library part from the example

std/std.solc
```diff
@@ -714,69 +750,14 @@ function getReader(d:ABIDecoder(ty, reader)) -> reader {
...
+    baseType : WordReader =>
+    instance ABIDecoder(calldata(DynArray(baseType)), CalldataWordReader):ABIDecode(calldata(DynArray(baseType_decoded)))
+{
+    function decode(ptr:ABIDecoder(calldata(DynArray(baseType)), reader), currentHeadOffset:word) -> calldata(DynArray(baseType)) {
+        let newptr = WordReader.advance(ptr, currentHeadOffset);
+        let reader: CalldataWordReader = getReader(newptr);
+        let addr: word = Typedef.rep(reader);
+        return Typedef.abs(addr);
+    }
+}
```
mbenke, line 776: Add an instance for `()`:

```suggestion
}

forall reader . reader:WordReader => instance ABIDecoder((), reader):ABIDecode(())
{
    function decode(ptr:ABIDecoder((), reader), currentHeadOffset:word) -> () {
      return ();
    }
}

```

mbenke:
It would be good to add one more method to the example.

PR #202, opened 2025-09-16
Fixes evm return type

rodrigogribeiro:


d-xo:
I had understood from the call today that @mbenke had specilization related concerns around the polymorphic return type?

mbenke:
Yes, but it's more about how it is used. Anyway, I agree that `return` and `revert`  should have the same type.

mbenke (approved):

PR #210, opened 2025-09-26
Sail objects

mbenke:
Add objects in SAIL (aka core) to reflect Yul objects, in preparation for adding constructor/deployer code.

Should be behaviour-neutral, except of the new output format:

```
object Answer {
  code {
    function main () -> word {
      return 42
    }
  }
}
```


rodrigogribeiro (approved):
Great!

PR #212, opened 2025-09-26
Contract deployment - stage 2: argumentless constructors

mbenke:


yule/Main.hs
```diff
@@ -21,45 +19,66 @@ main = do
...
-    deploycode = YulCode
-      [ calls "mstore" [yulInt 64, YCall "memoryguard" [yulInt 128]]
-      , ylva "memPtr" (YCall "mload" [yulInt 64])
-      -- call constructor here
-      , calls "datacopy" [yulInt 0, dataoffset, datasize]
-      , calls "return" [yulInt 0, datasize]
-      ]
+    calls f args = YExp (YCall f args)
+
+deployCode :: String -> Bool -> YulCode
```
d-xo, line 54: I'm wondering if this needs to be done during codegen or if it can also be a desugaring pass?

mbenke: Not sure what you mean - if mstore/rreturn above, then yes, it can be part of the dispatch code as we discussed; OTOH if you mean adding the deploy code, than it might be difficult as it is a different Yul subobject.

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -104,7 +106,9 @@ type Translation a = EM (a, [Core.Stmt])
 type CoreName = String
 
 emitTopDecl :: TopDecl Id -> EM [Core.Object]
-emitTopDecl (TContr c) = fmap pure (emitContract c)
+emitTopDecl (TContr c) = withLocalState do
```
d-xo, line 109: why do we need the `withLocalState` now?

src/Solcore/Desugarer/EmitCore.hs
```diff
@@ -120,18 +124,37 @@ emitContract :: Contract Id -> EM Core.Object
...
-    -- debug ["!! emitCDecl ", show cd]
     addData dt >> pure []
 emitCDecl cd = debug ["!! emitCDecl ", show cd] >> pure []
 
+-- By now, constructor should be converted to a function "constructor"
+
+findConstructor :: [ContractDecl Id] -> Maybe (FunDef Id)
+findConstructor = go where
+  go [] = Nothing
+  go (CFunDecl d:ds)| isConstructor d = Just d
```
d-xo:
```suggestion
  go (CFunDecl d:_)| isConstructor d = Just d
```

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -209,21 +210,51 @@ addDeclResolutions _ = return ()
...
-    forM_ entries (specEntry)
-    st <- gets specTable
-    dt <- gets spDataTable
-    let dataDecls = map (CDataDecl . snd) (Map.toList dt)
-    let funDecls = map (CFunDecl . snd) (Map.toList st)
-    let decls' = dataDecls ++ funDecls
-    return (TContr (Contract name args decls'))
+    -- Runtime code
+    runtimeDecls <- withLocalState do
+       forM_ entries (specEntry)
```
d-xo:
```suggestion
       forM_ entries specEntry
```

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -209,21 +210,51 @@ addDeclResolutions _ = return ()
...
+-- keep datatype defs intact
+specialiseTopDecl d@TDataDef{} = pure [d]
+-- Drop all toplevel decls that are not contracts - we do not need them anymore
+specialiseTopDecl decl = pure []
+
+findConstructor :: [ContractDecl Id] -> Maybe (Constructor Id)
+findConstructor = foldr (\d -> (getConstructor d <|>)) Nothing
+
+-- findConstructor (c:cs) = getConstructor c <|> findConstructor cs
+
```
d-xo:
```suggestion
```

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -209,21 +210,51 @@ addDeclResolutions _ = return ()
...
-    let funDecls = map (CFunDecl . snd) (Map.toList st)
-    let decls' = dataDecls ++ funDecls
-    return (TContr (Contract name args decls'))
+    -- Runtime code
+    runtimeDecls <- withLocalState do
+       forM_ entries (specEntry)
+       getSpecialisedDecls
+    -- Deployer code
+    modify (\st -> st { specTable = emptyTable })
+    let deployerName = Name (pretty name <> "$Deployer")
```
d-xo: Is this meant to be unused?

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -209,21 +210,51 @@ addDeclResolutions _ = return ()
...
-    let dataDecls = map (CDataDecl . snd) (Map.toList dt)
-    let funDecls = map (CFunDecl . snd) (Map.toList st)
-    let decls' = dataDecls ++ funDecls
-    return (TContr (Contract name args decls'))
+    -- Runtime code
+    runtimeDecls <- withLocalState do
+       forM_ entries (specEntry)
+       getSpecialisedDecls
+    -- Deployer code
+    modify (\st -> st { specTable = emptyTable })
```
d-xo, line 221: do we also need to clear the data table here?

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -209,21 +210,51 @@ addDeclResolutions _ = return ()
...
+    let deployerName = Name (pretty name <> "$Deployer")
+    let mconstructor = findConstructor decls
+    deployDecls <- case mconstructor of
+      Just c -> withLocalState do
+        cname' <- specConstructor c
+        st <- gets specTable
+        let cdecl = st Map.! cname'
```
d-xo: would prefer to avoid `!` and use something non partial, even we just `error` with a decent message on the not found branch.

d-xo: Is this supposed to be unused?

src/Solcore/Desugarer/Specialise.hs
```diff
@@ -237,6 +268,12 @@ specEntry name = withLocalState do
       Nothing -> do
         warns ["!! Warning: no resolution found for ", show name]
 
+specConstructor (Constructor [] body) = do
```
d-xo, line 267: type sigs for top level decls pls <3

yule/Main.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Main where
-import Language.Core(Object(..))
+-- import Language.Core(Object(..))
```
d-xo:
```suggestion
```

yule/Main.hs
d-xo: I mean that I think it would be nice to have something like an `init` that is considered an entrypoint by the specializer like `main` is right now, and we use the specialized contents of that for the `deployCode`. Remember we will need access to the stdlib and abi decoding in `deployCode` to implement constructor args, and imo it would also in general be more fitting with our "everything in language" philosophy.

mbenke: Yes, this is in the works, but that's what I start with.

In the upcoming PR, `deployCode` becomes much simpler:

```
deployCode _name withStart = YulCode $ go withStart where
  go True = [ [yulStmt| usr$start() |] ]
  go False = []
```

src/Solcore/Desugarer/EmitCore.hs
mbenke: So that if a compilation unit contains muiltiple contracts, code generation for each one is sandboxed.

src/Solcore/Desugarer/Specialise.hs
mbenke: Yes, so that the deployer code only contains functions it needs and not all functions used by the deployed contract.

src/Solcore/Desugarer/Specialise.hs
mbenke: not needed anymore, removed

src/Solcore/Desugarer/Specialise.hs
mbenke: not needed anymore, removed

mbenke: not needed anymore, removed

src/Solcore/Desugarer/Specialise.hs
mbenke: Done

PR #213, opened 2025-09-26
Desugar/ContractDispatch: init

d-xo:
This adds a desugaring pass that generates the standard linear solc style runtime entrypoint based on comparing the first four bytes of calldata against a computed selector hash. It does this using the data types and instances defined in std/dispatch.solc

Restrictions / Future Work:
- For now we assume all functions defined on a contract are public and non payable. Further work is required to introduce method modifiers into the surface syntax. Fallback and receive are not yet handled.
- Selector computation currently occurs at runtime, this requires the introduction of compile time evaluation to move to compile time.
- The current name type / instance generation approach should also hopefully be replaced by compile time / type level strings in the future.
- This PR also introduces that restriction that all top level contract methods must have a fully specified type signature.

Testing:

```
./runsol.sh std/dispatch.solc --calldata "add2(uint256,uint256)" 2 3
```

### DIFF
--- a/src/Solcore/Desugarer/ContractDispatch.hs
+++ b/src/Solcore/Desugarer/ContractDispatch.hs
@@ -41,22 +41,22 @@ genNameDecls (Contract cname _ cdecls) = foldl go Set.empty cdecls
     go acc _ = acc
 
 genMainFn :: Contract Name -> Contract Name
-genMainFn (Contract cname tys cdecls) = Contract cname tys ((CFunDecl mainfn) : cdecls)
+genMainFn (Contract cname tys cdecls) = Contract cname tys (CFunDecl mainfn : cdecls)
   where
-    mainfn = FunDef (Signature [] [] (Name "main") [] Nothing) body
+    mainfn = FunDef (Signature [] [] "main" [] Nothing) body
     body = [ StmtExp (Call Nothing (QualName "RunContract" "exec") [cdata])]
-    cdata = Con (Name "Contract") [methods, fallback]
+    cdata = Con "Contract" [methods, fallback]
     methods = tupleExpFromList (fmap mkMethod (mapMaybe unwrapSigs cdecls))
-    fallback = Con (Name "Fallback")
-      [ proxyExp (TyCon (Name "NonPayable") [])
+    fallback = Con "Fallback"
+      [ proxyExp (TyCon "NonPayable" [])
       , proxyExp unit
       , proxyExp unit
-      , Var (Name "revert_handler")
+      , Var "revert_handler"
       ]
 
-    mkMethod (Signature _ _ fname fargs (Just ret)) | all isTyped fargs = Con (Name "Method")
+    mkMethod (Signature _ _ fname fargs (Just ret)) | all isTyped fargs = Con "Method"
       [ proxyExp (TyCon (nameTypeName cname fname) [])
-      , proxyExp (TyCon (Name "NonPayable") [])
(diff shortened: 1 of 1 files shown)


src/Solcore/Primitives/Primitives.hs
```diff
@@ -154,6 +154,20 @@ primTrue = (trueName, monotype boolTy)
...
 
+-- tuple utils
+
+tupleExpFromList :: [Exp Name] -> Exp Name
+tupleExpFromList [] = Con (Name "()") []
+tupleExpFromList [e] = e
+tupleExpFromList [e1,e2] = epair e1 e2
+tupleExpFromList (e1 : es) = epair e1 (tupleExpFromList es)
+
+tupleTyFromList :: [Ty] -> Ty
```
mbenke, line 165: This seems to duplicate `mkTupleTy` in `SolcoreParser`

src/Solcore/Primitives/Primitives.hs
```diff
@@ -154,6 +154,20 @@ primTrue = (trueName, monotype boolTy)
 primFalse :: (Name, Scheme)
 primFalse = (falseName, monotype boolTy)
 
+-- tuple utils
+
+tupleExpFromList :: [Exp Name] -> Exp Name
```
mbenke, line 159: This seems to duplicate `tupleExp` in `SolcoreParser`

d-xo:
Some questions:

1. what should happen if a `main` is already defined?
2. should we move the requirement for fully annotated types on contract methods to the parser?
3. should the desugaring pass automatically import `dispatch`?

mbenke:
Ad 1. If `main` is already defined, we might skip generating dispatch. This would allow users to provide their own dispatch code, should their feel so inclined.
Ad 2. This feels like part of a bigger discussion (e.g. should we allow `function foo() {}`)
Ad 3. Again, this should be handled more systematically - we should have something akin to Haskell `Prelude` (possibly with `--no-implicit-prelude` flag), including dispatch, assignment and operator desugaring

src/Solcore/Desugarer/ContractDispatch.hs
```diff
@@ -0,0 +1,156 @@
...
+    go acc (CFunDecl (FunDef sig _)) = let
+        dataTy = mkNameTy cname (sigName sig)
+        instDef = mkNameInst dataTy (sigName sig)
+      in Set.union (Set.fromList [TDataDef dataTy, TInstDef instDef]) acc
+    go acc _ = acc
+
+genMainFn :: Contract Name -> Contract Name
+genMainFn (Contract cname tys cdecls) = Contract cname tys ((CFunDecl mainfn) : cdecls)
+  where
+    mainfn = FunDef (Signature [] [] (Name "main") [] Nothing) body
```
mbenke: `Name` is an instance of `IsString` so instead of `(Name "main")`, you can write just `"main"` - but of course it's a matter of taste

mbenke (approved):
Well done, just a couple of nits.

src/Solcore/Primitives/Primitives.hs
d-xo: That is a different `Ty` from `SyntaxTree`, not the one from `Syntax` that we use here.

src/Solcore/Primitives/Primitives.hs
d-xo: That is a different `Exp` from `SyntaxTree`, not the one from `Syntax` that we use here.

PR #218, opened 2025-09-30
runsol: interactive debugging & enhanced returndata decoding

d-xo:
- uses geth instead of hevm for evm execution
- install geth and goevmlab into the nix devshell
- remove hevm from the nix devshell
- bump our nixpkgs commit to the latest
- `runsol.sh` now abi decodes returndata based on the provided calldata sig
- `runsol.sh` has a new --debug flag that will open an interactive evm debugger (traceview from (goevmlab)[https://github.com/holiman/goevmlab] for the current execution

A screenshot of the interactive debugger:

image

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,22 +1,5 @@
 {
   "nodes": {
-    "ethereum-tests": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1698656397,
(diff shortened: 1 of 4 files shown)


rodrigogribeiro (approved):
LGTM.

PR #223, opened 2025-10-01
Fixes the issue on polymorphic types after pattern matching compilation

### LINKED ISSUE #219: Too much polymorphism in the match compiler
In the program

```
forall a b . function fst(p: (a, b)) -> a {
    match p {
    | (a, _) => return a;
    }
}
contract TestUnitMatch {
  function main() -> () {
    match ((), ()) {
     | x => return fst(x);
    }
  }  
}
```

The mctch compiler invents too many type variables in `fst`:

```
> Match compilation result:

forall a b . function fst (p<(a, b)> : (a, b)) -> a {
   match (p<(d, b)>) {
   | pair<d -> b -> (d, b)>(var_3<d>, var_4<var_1>) =>
      let var_5<d> = var_3<d> ;
      let var_7<var_1> = var_4<var_1> ;
      return var_5<d>;
   }
}
contract TestUnitMatch {
   function main () -> () {
      let var_1<((), ())> : ((), ()) = pair<() -> () -> ((), ())>(()<()>, ()<()>) ;
      return fst<((), ()) -> ()>(var_1<((), ())>);
   }
}
```

which leads to a specialisation failure which should be avoidable if these invented variables were unified with a and b:

```
PANIC: spec(let var_5<d> = var_3<d> ;): free type vars in d: [TVar {var = d}] @ subst={_a |-> (), _b |-> ()}
```

rodrigogribeiro:
Fixes #219 by moving wildcard desugaring to happen before type inference.

### DIFF
--- a/src/Solcore/Desugarer/MatchCompiler.hs
+++ b/src/Solcore/Desugarer/MatchCompiler.hs
@@ -13,7 +13,6 @@ import qualified Data.List.NonEmpty as L
 
 import Language.Yul
 
-import Solcore.Desugarer.ReplaceWildcard
 import Solcore.Frontend.Pretty.SolcorePretty
 import Solcore.Frontend.Syntax
 import Solcore.Frontend.TypeInference.Id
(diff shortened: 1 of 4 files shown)


src/Solcore/Desugarer/MatchCompiler.hs
```diff
@@ -588,4 +585,42 @@ eqnsType :: Equations Id -> Ty
...
+inc = do
+  i <- get
+  put (i + 1)
+  return i
+
+freshName :: CompilerM Name
+freshName
+  = do
+        n <- inc
+        -- pre <- ask
```
d-xo:
```suggestion
```

d-xo (approved):
a tiny tiny nit, but otherwise looks great

PR #231, opened 2025-10-07
Yul quasiquoters

mbenke:
Quasiquoters for Yul: 

Antiquotation is supported (variables only) using bacticks or `${ }`.
Example use:

``` haskell
deployCode name withConstructor = YulCode $ [yulBlock|
  {
    mstore(64, memoryguard(128))
    let memPtr := mload(64)
  }
  |]
  <> callConstructor withConstructor
  <> [yulBlock|
     { datacopy(0, `dataoffset`, `datasize`)
       return(0, `datasize`)
     } |]
  where
    cname = yulString name
    callConstructor True = pure [yulStmt| usr$constructor() |]
    callConstructor False = []
    datasize = [yulExp| datasize(${cname}) |]
    dataoffset = [yulExp| dataoffset(`cname`) |]
```

quasiquoters for patterns and expressions:

``` haskell
simplify [yulExp|add(2,2)|] = [yulExp| 4|]
```

### DIFF
--- a/src/Language/Yul/QuasiQuote.hs
+++ b/src/Language/Yul/QuasiQuote.hs
@@ -19,21 +19,27 @@ yulExp :: QuasiQuoter
 yulExp = simpleQuoterUsing Parser.yulExp
 
 simpleQuoterUsing :: Data a => Parser a -> QuasiQuoter
-simpleQuoterUsing = simpleQuoter . quoteUsingParser
+simpleQuoterUsing p = simpleQuoter (quoteUsingLift liftYulExp p) (quoteUsingLift liftYulPat p)
 
-simpleQuoter :: (String -> Q Exp) -> QuasiQuoter
-simpleQuoter q = QuasiQuoter
-  { quoteExp = q
-  , quotePat = undefined
-  , quoteDec = undefined
-  , quoteType = undefined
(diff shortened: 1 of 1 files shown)


src/Language/Yul/QuasiQuote.hs
```diff
@@ -0,0 +1,45 @@
...
+
+simpleQuoterUsing :: Data a => Parser a -> QuasiQuoter
+simpleQuoterUsing = simpleQuoter . quoteUsingParser
+
+simpleQuoter :: (String -> Q Exp) -> QuasiQuoter
+simpleQuoter q = QuasiQuoter
+  { quoteExp = q
+  , quotePat = undefined
+  , quoteDec = undefined
+  , quoteType = undefined
```
d-xo: Not entirely sure what these are for, but maybe at least an `error` with a message just to make potential future debugging a bit friendlier?

mbenke: This being a TH, it would be a compile error really, and who would use Yul code in **types**? But at least you have given me an excuse to procrastinate strings (*cringe*) and write a quasiquoter for patterns ;)

mbenke: added a quasiquoter for patterns, so now you can write

``` haskell
simplify [yulExp|add(2,2)|] = [yulExp|4|]
```

PR #243, opened 2025-10-17
Storable class and copying strings between memory and storage

mbenke:


std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
     function assign(l:lhs, r:rhs) -> ();
 }
 
+// a can be stored in storage(b); e.g. memory(string) : Storable(string)
+pragma no-coverage-condition Storable, Assign;
+forall a b.
+class a:Storable(b) {
```
d-xo, line 1261: when would `b` need to be different from `a`

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+      }
+  }
+}
+
+
+// shamelessly stolen from abi_encode_t_string_storage_to_t_string_memory_ptr
+function loadBytesFromStorage(slot:word, memPtr:word) -> word {
+  let ret : word;
+  assembly {
+    let pos := memPtr
```
d-xo, line 1345: seems superfluous?

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+
+// Shamelessly stolen from  function copy_byte_array_to_storage_from_t_bytes_memory_ptr_to_t_bytes_storage
+// TODO: check https://notes.argot.org/@solidity-storage-wrap-behavior-audit#Iteration-over-array-items
+function storeBytesFromMemory(slot: word, src: word) -> () {
+  assembly {
+    let newLen := mload(src)
+    // TODO: check old len, cleanup etc
+    let srcOffset := 32
+    switch gt(newLen, 31)
+      case 1 {
```
d-xo, line 1308: I would find this a bit easier to parse if we indented the following block

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+	  if lt(loopEnd, newLen) {
+	      let lastValue := mload(add(src, srcOffset))
+	      let lastLen := and(newLen, 0x1f)
+	      let mask := not(shr(mul(8, lastLen), not(0)))
+	      let nudata := and(lastValue, mask)  // a Yul variable cannot be called "data". Go figure.
+	      sstore(dstPtr, nudata)
+	  }
+	  sstore(slot, add(mul(newLen, 2), 1))
+      }
+      default {
```
d-xo, line 1327: same here re: indentation

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+	  let nudata := and(value, mask)
+	  let used := or(nudata, mul(2, newLen))
+	  sstore(slot,used)
+      }
+  }
+}
+
+
+// shamelessly stolen from abi_encode_t_string_storage_to_t_string_memory_ptr
+function loadBytesFromStorage(slot:word, memPtr:word) -> word {
```
d-xo, line 1342: This function would also imo benefit from a bit more indentation throughout

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+    let pos := memPtr
+    let slotValue := sload(slot)
+    let length := div(slotValue, 2)
+    let outOfPlaceEncoding := and(slotValue, 1)
+    if iszero(outOfPlaceEncoding) {
+	length := and(length, 0x7f)
+    }
+    mstore(pos, length)
+    pos := add(pos, 0x20)
+    switch and(slotValue, 1)
```
d-xo: I think we can just use `outOfPlaceEncoding` here? since we have it already?

std/std.solc
```diff
@@ -1246,14 +1255,130 @@ class lhs:Assign(rhs) {
...
+
+    case 0 {
+	// short byte array
+	mstore(pos, and(slotValue, not(0xff)))
+	//
+	// ret := add(pos, mul(0x20, iszero(iszero(length))))
+	// let notzero := iszero(iszero(length))
```
d-xo:
```suggestion
```

std/std.solc
mbenke: See thge comment above: memory(string) : Storable(string)

std/std.solc
mbenke: Fixed - it were just some stray tabs

std/std.solc
d-xo: ah of course I can't read :see_no_evil:

std/std.solc
mbenke: Fixed

PR #248, opened 2025-10-29
Fixes invokable generation and closure conversion for type annotated functions

rodrigogribeiro:
* Adds a new desugaring step which replaces function type annotated arguments by invokable constraints before type inference.
* Fixes a bug in closure conversion in the presence of type annotations.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcSimplify.hs
+++ b/src/Solcore/Frontend/TypeInference/TcSimplify.hs
@@ -82,7 +82,7 @@ checkEntails qs rs
       info [">>! Trying to check the entailment of:", pretty rs, " from:", pretty qs]
       let
           qs' = nub $ concatMap (bySuperM ctable) qs
-          unsolved q = not (isInvoke q) && not (entail ctable itable qs' q)
+          unsolved q = not (isHnf q) && not (isInvoke q) && not (entail ctable itable qs' q)
           -- compiler generated instances can introduce invokable constraints
           -- not present in the called function. Since type inference can produce
           -- such constraints, we do not consider them here.


src/Solcore/Desugarer/ReplaceFunTypeArgs.hs
```diff
@@ -0,0 +1,46 @@
...
+                , sigContext = concat pss ++ sigContext sig 
+                , sigParams = args'
+                }
+
+
+freshTy :: ReplaceM (Ty, Tyvar) 
+freshTy = do 
+  i <- get 
+  let
+    v = TVar $ Name ("v_" ++ show i)
```
d-xo: do we want to handle potential name collisions with user defined types here?

src/Solcore/Frontend/TypeInference/TcSimplify.hs
```diff
@@ -39,13 +39,40 @@ reduce qs0 ps0
...
       info ["<<!! Unsolved:", pretty unsolved]
       unless (null unsolved) (unsolvedError unsolved)
       pure rs
 
+enforceDependencies :: [Pred] -> TcM [Pred]
+enforceDependencies ps 
+  = do
+      let 
+        eqMainTy p p' = predMain p == predMain p' 
+        pss = splits eqMainTy ps
```
d-xo, line 56: will this group instances of different classes that have the same main type together? do we want the weak types to unify if they are instances of different classes?

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -244,7 +244,7 @@ tcExp e@(Lam args bd _)
          (e, t) <- closureConversion vs (apply s args') (apply s bd') ps1 ty
          -- here we do not need the constraints in ps1, since after closure conversion
          -- the lambda is replaced by a unique type constructor for it.
```
d-xo: comment seems to be out of date now?

src/Solcore/Desugarer/ReplaceFunTypeArgs.hs
rodrigogribeiro: Fixed by prefixing names with "$", avoiding clashing.

src/Solcore/Frontend/TypeInference/TcSimplify.hs
rodrigogribeiro: Thanks. Now it is fixed by considering the class name in constraints for this comparison.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: Since these comments are not relevant anymore, I simply removed them.

d-xo (approved):

PR #259, opened 2025-11-03
Adds arity checking on constraints, function signatures and instances.

rodrigogribeiro:


test/Cases.hs
```diff
@@ -232,6 +232,18 @@ cases =
...
     , runTestForFile "yul-return.solc" caseFolder
+    , runTestForFile "pragma_merge_base.solc" caseFolder
+    , runTestForFile "pragma_merge_import.solc" caseFolder
+    , runTestForFile "pragma_merge_verify.solc" caseFolder
+    , runTestExpectingFailure "pragma_merge_fail_patterson.solc" caseFolder
+    , runTestExpectingFailure "pragma_merge_fail_coverage.solc" caseFolder
+    , runTestForFile "single-lambda.solc" caseFolder
+    , runTestExpectingFailure "duplicated-type-name.solc" caseFolder
+    , runTestExpectingFailure "overlapping-heads.solc" caseFolder
+    , runTestExpectingFailure "instance-wrong-sig.solc" caseFolder
```
mbenke, line 243: instance-wrong-sig fails for the wrong reasons:
- lexical error
- undefined `Proxy`

PR #262, opened 2025-11-05
IfDesugarer: flip order of true / false constructors

d-xo:
Flips the desugaring of `true` / `false` so they compile down to `1`/`0` in yul as expected.

### DIFF
--- a/src/Solcore/Desugarer/IfDesugarer.hs
+++ b/src/Solcore/Desugarer/IfDesugarer.hs
@@ -34,8 +34,11 @@ desugarStmt v = v
 -- desugaring boolean data constructors
 
 desugarBoolCons :: Exp Id -> Exp Id
-desugarBoolCons (Con c es)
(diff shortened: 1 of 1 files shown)


src/Solcore/Desugarer/IfDesugarer.hs
```diff
@@ -34,8 +34,11 @@ desugarStmt v = v
 -- desugaring boolean data constructors
 
 desugarBoolCons :: Exp Id -> Exp Id
-desugarBoolCons (Con c es)
-  | isBoolCon c = sumConsFor c
+desugarBoolCons (Con c@(Id n _) es)
```
d-xo, line 37: I inlined this (and `desugarBoolPat`) to squash a warning about non-exhaustive pattern matching.

mbenke (approved):
Nice!

rodrigogribeiro (approved):
Great!

PR #271, opened 2025-11-10
blog: general wording tweaks

d-xo:
I removed the `CanRefund` example because I decided it was a bit semantically weird and didn't match our `processAuction` function. Otherwise just a bunch of wording tweaks.


rodrigogribeiro (approved):
Thanks!

PR #276, opened 2025-11-28
Contract testing using testrunner and evmone

mbenke:
Contract testing (create contract and call its methods) using testrunner and evmone. 
This PR integrates the C++ testrunner into the Nix build system and adds end-to-end integration tests to `nix flake check` (which is now run as part of the CI workflow).

Written using substantial Claude Code input, hence includes CLAUDE.md (which can also serve as a part of documentation).

###  Nix Build System:
  - Added Nix derivations for evmone dependencies (nix/evmone.nix, nix/intx.nix, nix/blst.nix)
  - Created testrunner package that builds the C++ test executor
  - Added checks.contests derivation that runs full pipeline integration tests
  - Uses nixpkgs packages (boost, nlohmann_json) instead of git submodules

###  Script Portability:
  - Made contest.sh configurable via environment variables with sensible defaults
  - Scripts work both in Nix builds (with overrides) and local development (with defaults)
  - Environment variables: SOLCORE_CMD, YULE_CMD, testrunner_exe, evmone

###  Testing:
  - Integration tests verify full pipeline: .solc → sol-core → .core → yule → .yul → solc → .hex → testrunner
  - nix flake check now runs contest test suite (6 tests)
  - Test results: basic.json (4 tests) and neg.json (2 tests) passing ✓

###  Documentation:
  - Added CLAUDE.md with architecture and Nix integration details
  - Updated README.md with testrunner build and usage instructions
  - Clear distinction between Haskell unit tests and integration tests

###  Usage

```
  # Run integration tests via Nix (recommended)
  nix flake check

  # Or build testrunner manually and run locally
  cmake -S . -B build && cmake --build build --target testrunner
  export testrunner_exe=build/test/testrunner/testrunner
  bash run_contests.sh
```


run_contests.sh
```diff
@@ -0,0 +1,3 @@
+bash ./contest.sh test/examples/dispatch/basic.json
```
Y-Nak, line 2:
```suggestion
set -euo pipefail
bash ./contest.sh test/examples/dispatch/basic.json
```
It's necessary to add `set -e` option at least. Otherwise, if the final test succeeds, errors in preceding tests may not be detected.

contest.sh
```diff
@@ -0,0 +1,82 @@
...
+file=$1
+shift
+
+if [[ ! -f "$file" ]]; then
+    echo "Error: File '$file' not found"
+    exit 1
+fi
+
+echo "Processing: $file"
+root_dir="$(cd "$(dirname "$(readlink --canonicalize "${BASH_SOURCE[0]}")")" && pwd)"
```
Y-Nak, line 23:
```suggestion
root_dir="$(cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")" && pwd)"
```
`readlink -f` doesn't work in macos(if `coreutils` are not installed). Probably `realpath` would be more portable.

Y-Nak (commented):
I added two nits.
Also, currently, the tests fail because of the `copyToMem` implementation. But other than that, LGTM overall.

CLAUDE.md
clonker, line 1: not sure how this is relevant to the pr, perhaps split it into a separate one?

CLAUDE.md
```diff
@@ -0,0 +1,366 @@
...
+- `evmone` - Path to evmone library (default: `"~/.local/lib/libevmone.so"`)
+
+### Nix Integration
+
+The project uses Nix flakes for reproducible builds:
+
+**Packages** (`nix build .#<package>`):
+- `sol-core` - Main Haskell compiler
+- `testrunner` - C++ testrunner binary
+- `intx`, `blst`, `evmone` - EVM dependencies (built from source)
```
clonker, line 295: why though? intx and blst are pulled by evmone's cmake in a pinned version - the supported and tested combination, I would assume. Can't we just rely on that instead of potentially introducing version drift?

clonker: also i'm not sure what happens with `nlohmann_json` here, is Hunter completely disabled, which version does it take?

contest.sh
mbenke: > `readlink -f` doesn't work in macos(if `coreutils` are not installed). Probably `realpath` would be more portable.

Does `realpath` work on macos without `coreutils`? On my (admittedly ancient - macOS 10.13) MacBook it is the oposite - system `readlink` and `realpath` from coreutils.
Is readlink an issue for you even when using `nix develop`? 

We can always bemore paranoid and try
```
root_dir="$(realpath "$(dirname "${BASH_SOURCE[0]}")" || cd "$(dirname "$(readlink --canonicalize "${BASH_SOURCE[0]}")")" && pwd)"
```

CLAUDE.md
mbenke: @clonker I am no expert on Nix, but from what I am told, in Nix builds, you're in a sandbox with no internet access. So Hunter can't download packages. That's why the current evmone.nix:                                      
- Disables Hunter completely (lines 24-32) by stubbing out the Hunter macros          
- Provides intx and blst explicitly from Nix instead

Can you suggest a better approach that works with Nix?

clonker: I can't, I'm not a Nix user at all; it just struck me as odd and potentially unstable / hard to maintain.

flake.nix
```diff
@@ -27,25 +27,102 @@
...
+
+            checkPhase = ''
+              cd ..
+              export PATH=${sol-core}/bin:${pkgs.solc}/bin:${pkgs.jq}/bin:$PATH
+
+              # Override commands and paths to use Nix-provided binaries
+              export SOLCORE_CMD="sol-core"
+              export YULE_CMD="yule"
+              export testrunner_exe=build/test/testrunner/testrunner
+              export evmone=${evmone-lib}/lib/libevmone.so
```
Y-Nak:
```suggestion
              export evmone=${evmone-lib}/lib/libevmone${pkgs.stdenv.hostPlatform.extensions.sharedLibrary}
```
This would a more portable one and is actually necessary to run on macOS.

Y-Nak (commented):
I suggested one change to be able to run `nix flake check` on macos.

contest.sh
Y-Nak: `realpath` works on macos without `coreutils` (probably since around 10.15?). But `nix develop` and `nix flake check` works well with `readlink --canonical`. So it'd be fine to leave it as is for now.

CLAUDE.md
clonker: Well I think it's fine. We just shouldn't forget to update intx and blst together with evmone. I'm not sure about the specifics but perhaps we can encapsulate the intx, blst, and nlohmann_json deps in the evmone package so they're not top-level? They're not runtime dependencies but statically included in the binary.

mbenke:
Superseded by #314 and #323

PR #282, opened 2025-12-05
Fix dependency analysis to consider class names as declarations

rodrigogribeiro:


std/std.solc
```diff
@@ -152,7 +152,7 @@ function ge(x:a, y:a) -> bool {
 
 forall a. a:Ord =>
 function lt(x:a, y:a) -> bool {
-  return gt(y,x);
```
mbenke, line 155: Why is this needed? Note that there is also a function gt:

```
forall a. a:Ord =>
function gt(x:a, y:a) -> bool {
  return not(Ord.gt(x,y));
}
```

This change might also break desugaring of `>`:

```
  elab (S.ExpGT e1 e2) = do
     (e1', e2') <- elab (e1, e2)
     let fun = QualName (Name "Num") "gt"
     pure $ Call Nothing fun [e1', e2']
```

... if it were not already broken

mbenke: Fixing comparison desugaring is planned in #284

rodrigogribeiro: I changed this definition because I thought it was a typo. But it helped me to uncover another bug in dependency analysis. Now, with the latest commit, it is working fine.

Issue #226, opened 2025-10-07
Modify the parser to enforce fully annotated types for contract & instance method


d-xo:
https://github.com/argotorg/solcore/blob/3710dcf663168a4505ce5280eda02f42fc09bb54/src/Solcore/Desugarer/ContractDispatch.hs#L64

PR #291, opened 2025-12-12
Makes a parser error when instance and contract functions aren't full…

rodrigogribeiro:
This PR solves issue #226 by modifying the parser to only accept instance and contract functions which are fully type annotated. Almost all changes are due to make all test cases compile due to the parser change.


rodrigogribeiro:
Closing in favor of #379.

PR #292, opened 2025-12-15
Checking if type variables are unbound in class and instance definitions

### LINKED ISSUE #240: foralls not enforced/undefined types not caught in class instances
```
forall self.
class self:C {
    function size(x:self) -> word;
}

instance ():C {
    function size(x:()) -> word {
        return 0;
    }
}

instance uint:C {
    function size(x:uint) -> word {
        return 1;
    }
}
```
Rather than complain about `uint, as it should, the typechecker issues a head-scratching error:

```
Overlapping instances are not supported
instance:
?$3 : C
overlaps with:
() : C

 - in:instance uint : C {
   function size (x : uint) -> word {
      return 1;
   }
}
```

rodrigogribeiro:
Fixes #240.


test/examples/cases/unbound-instance-var.solc
mbenke, line 1: The file  name wrong-overlapping.solc suggests it's testing overlapping instances, but it's actually testing unbound type variables. Perhaps rename to something like:
  - unbound-type-var-in-instance.solc
  - missing-forall-quantifier.solc
  - undeclared-instance-var.solc

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -803,6 +803,10 @@ checkConstraints = mapM_ checkConstraint
 checkInstance :: Instance Name -> TcM ()
 checkInstance idef@(Instance d vs ctx n ts t funs)
   = do
+      -- checking if all variables are declared
+      let unbound_vars = (bv idef) \\ vs
+      unless (null unbound_vars) $ do
+        unboundTypeVars idef unbound_vars
```
mbenke: Code duplication with the similar change in TcContract and also in checkSignature. Consider

```
  checkAllTypeVarsBound :: Pretty a => a -> [Tyvar] -> [Tyvar] -> TcM ()
  checkAllTypeVarsBound context used declared =
    let unbound = used \\ declared
    in unless (null unbound) $ unboundTypeVars context unbound
```
  Then use:
  -- In checkClass:

```
  checkAllTypeVarsBound icls (v : vs) bvs

  -- In checkInstance:
  checkAllTypeVarsBound idef (bv idef) vs

  -- In checkSignature:
  checkAllTypeVarsBound sig (v : vs) bvs
```

test/examples/cases/unbound-instance-var.solc
rodrigogribeiro: Renamed to `unbound-instance-var.solc`.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: Fixed in the latest commit. Thanks for the suggestion!

Issue #264, opened 2025-11-05
String literals


mbenke:
Idea:

```
import dispatch;
// function f() -> memory(string) { return "uint256"; }) should desugar to

function stringLit75696e74323536() -> memory(string) {
        let size : word = 7;
	let ptr = allocate_memory(64);
        assembly {
            mstore(ptr, size)
            mstore(add(ptr,32), 0x75696e7432353600000000000000000000000000000000000000000000000000)
        }
	return memory(ptr);
}

function f() -> memory(string) {
  return stringLit75696e74323536();
}
```

with appropriate chunking for longer strings.

Parts of #176 can be reused.

alexoltean61:
Hi, nice project! I've been lurking around for the past week or so, reading the [blog post](https://www.soliditylang.org/blog/2025/11/14/core-solidity-deep-dive/) and playing with the standard library.

I was wondering, are contributions welcome? For me it would be an opportunity to learn more about compilers and type theory, and get to the guts of the EVM.

E.g., I had a bit of free time today and gave this issue a go, following @mbenke's sketch and his older code. I made a fork with my changes [over here](https://github.com/alexoltean61/solcore/tree/alexoltean61/string-literals). Would it be okay if I turned it into a PR?

PR #294, opened 2025-12-22
String literal support

alexoltean61:
This PR adds basic support for string literals, implementing #264. Each string literal is desugared to the application of a new function, which stores chunks in memory and returns their starting pointer.

In the test cases, I used this to define high-level `revert` and `assert` functions similar to regular Solidity.

Please note that string literals in pattern matching equations are still **not** supported. For example, this code snippet will not compile (but it does not compile on the main branch either):
```solidity
import dispatch;

contract Match {
    function matchStr() -> () {
        match "test" {
            | "test" => return ();
            | _ => return ();
        }
    }
}
```


mbenke:
Obsoleted by comptime string handling mechanisms

PR #299, opened 2025-12-30
Fixes the free variable problem on closure conversion

rodrigogribeiro:
* Fixes the free variable problem on closure conversion
* Adds test cases.


src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1111,45 +1111,60 @@ mword = monotype word
 -- determining free variables
 
 class Vars a where
-  vars :: a -> [Id]
+  free :: a -> [Id]
+  bound :: a -> [Id]
 
 instance Vars a => Vars [a] where
-  vars = foldr (union . vars) []
+  free = foldr (union . free) []
```
mbenke: This may be slightly misleading, for a statement list like `[let y = 42; return y]`, the free will include y even though it's bound locally. It is compensated for in `closureConversion` but I think it merits at least a comment in the class definition.

This approach also may result in incorrect closure - see comment about `If`  statement

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1111,45 +1111,60 @@ mword = monotype word
...
+instance Vars (Equation Id) where
+  free (ps, ss) = free ss \\ bound ps
+  bound _ = []
+
+instance Vars (Exp Id) where
+  free (Var n) = free n
+  free (Con _ es) = free es
+  free (FieldAccess Nothing _) = []
+  free (FieldAccess (Just e) _) = free e
+  free (Call (Just e) n es) = free e `union` free n `union` free (e : es)
```
```suggestion
  free (Call (Just e) n es) = free e `union` free n `union` free es
```

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -289,10 +289,10 @@ closureConversion vs args bdy ps ty
       sch <- generalize (ps', ty)
       let
           fn = Name $ "lambda_impl" ++ show i
-          argsn = map idName (vars args)
+          argsn = map idName $ (bound args) ++ (bound bdy)
```
```suggestion
          argsn = map idName $ bound args ++ bound bdy
```

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -289,10 +289,10 @@ closureConversion vs args bdy ps ty
...
           defs = fs ++ argsn ++ Map.keys primCtx
-          free = filter (\ x -> notElem (idName x) defs) (vars bdy)
-      if null free then do
+          freevs = filter (\ x -> notElem (idName x) defs) (free bdy)
```
mbenke: This may be more readable
```suggestion
          freevs = [x | x <- free bdy, idName x `notElem` defs]
```

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1111,45 +1111,60 @@ mword = monotype word
...
+  bound (Untyped n) = [n]
+
+instance Vars (Stmt Id) where
+  free (e1 := e2) = free [e1,e2]
+  free (Let _ _ (Just e)) = free e
+  free (Let _ _ _) = []
+  free (StmtExp e) = free e
+  free (Return e) = free e
+  free (Match e eqns) = free e `union` free eqns
+  free (If e blk1 blk2) = free e `union` free blk1 `union` free blk2
```
mbenke, line 1152: For `let y:word;if b then { let z = 7; y := z; } else ...` we have `z` in `free` and not in `bound`. Then closure conversion tries to close over `z`.

PR #301, opened 2026-01-02
Specialiser  refactoring

mbenke:


src/Solcore/Backend/TVSubst.hs
Y-Nak, line 1: It seems this file is a dead file.

src/Solcore/Backend/SpecMonad.hs
```diff
@@ -0,0 +1,401 @@
...
+-- >>> r2 <> r2
+-- TVR {unTVR = [(TVar {var = a},TVar {var = c}),(TVar {var = b},TVar {var = a}),(TVar {var = c},TVar {var = b})]}
+-- >>> r2 <> r2 <> r2
+-- TVR {unTVR = []}
+
+instance Semigroup TVRenaming where
+  r1 <> r2 = TVR (filter (uncurry (/=)) [ (u, renameTV r1 v) | (u, v) <- unTVR r2])
+
+instance Monoid TVRenaming where
+  mempty = TVR mempty
```
Y-Nak, line 388: It seems `<>` implementation doesn't satisfy the monoid law here. The right identity law is violated,
e.g., `TVR [(a, b)] <> TVR []` should be equal to `TVR [(a, b)]`, but actually returns `TVR []`.

src/Solcore/Backend/TVSubst.hs
mbenke: Indeed. I tried to make it a separate module but failed due to cyclic imports. Removed.

src/Solcore/Backend/SpecMonad.hs
mbenke: Good catch! I have focused so much on removing redundant pairs, I forgot about this. Fixed.

Y-Nak (approved):
Great!

PR #308, opened 2026-01-13
Fixes the closure typing in instance definitions.

rodrigogribeiro:


rodrigogribeiro:
Closing in favor of #313, which deals with it in a more general way.

PR #310, opened 2026-01-14
Fixes instance checking

rodrigogribeiro:
* This PR adds some checks for instance type inference: 
1.  Kind check all instance type arguments; 
2.  Checks class arity for instances and 
3.  Checks if all constraints have the correct arity.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -1206,8 +1206,8 @@ rename t = let vs = bv t
 
 classArityError :: Pretty a => Name -> ClassInfo -> a -> TcM ()
 classArityError n cinfo v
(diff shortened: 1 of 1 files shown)


src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1194,6 +1204,14 @@ rename t = let vs = bv t
 
 -- errors
 
+classArityError :: Pretty a => Name -> ClassInfo -> a -> TcM ()
+classArityError n cinfo v
+  = throwError $ unlines [ "Type class:" ++ pretty n
+                         , "has arity of:" ++ show (classArity cinfo)
+                         , "which does not match:"
+                         , pretty v
```
```suggestion
  = throwError $ unlines [ "Type class " ++ pretty n
                         , "requires " ++ show (classArity cinfo) ++ " weak parameter(s)"
                         , "which does not match:"
                         , pretty v
```

PR #311, opened 2026-01-14
Makes incomplete instances an error.

### LINKED ISSUE #307: Disallow incomplete instance definitions
Currently, the compiler allows class instances to omit method implementations. We have decided to change this behavior and no longer allow instances with unimplemented methods.
After this change, all methods required by a class/trait must be fully implemented in its instances, and missing implementations should result in a compile-time error.

rodrigogribeiro:
This PR fixes #307.


mbenke (approved):
LGTM

PR #313, opened 2026-01-16
Fixes instance checking and elaboration for members which return closures

### LINKED ISSUE #302: Closure types don't unify with function types in instance methods
The following compiles:
```solidity
function ct(x : word) -> ((word) -> word) {
    return lam(y : word) {
        return x;
    };
}
```
The following does not:
```solidity
forall t . class t:CtFun {
    function ct(x : t) -> ((t) -> t);
}

instance word:CtFun {
    function ct(x : word) -> ((word) -> word) {
        return lam(y : word) {
            return x;
        };
    }
}
```
Error:
```
The instance method:
CtFun.ct
has the following infered type:
word -> word -> word
which is not an valid instance for:
word -> t_closure17

 - in:instance word : CtFun {
   function CtFun.ct (x<word> : word) -> t_closure17 {
      return t_closure17<word -> t_closure17>(x<word>);
   }
}
```

rodrigogribeiro:
This PR fixes #302 . 

The strategy is to type the code twice: first we run the type inference without any desugaring. This is necessary becase code with closures will change its type. Such desugaring strategy is a problematic in the presence of type signatures, since the infered type for the desugared term will not match the annotated type. The strategy is detect all possible errors in the first step and then desugar and annotate all terms in the second one.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcContract.hs
+++ b/src/Solcore/Frontend/TypeInference/TcContract.hs
@@ -29,18 +29,14 @@ import Solcore.Primitives.Primitives
 typeInfer :: Option ->
              CompUnit Name ->
              IO (Either String (CompUnit Id, TcEnv))
-typeInfer options c@(CompUnit imps decls)
+typeInfer options (CompUnit imps decls)
   = do
-      r <- runTcM (tcCompUnit c) (initTcEnv noDesugarOpt)
+      r <- runTcM (tcCompUnit (CompUnit imps decls)) (initTcEnv options)
       case r of
-        Left err -> pure $ Left err
-        Right _ -> do
-          r <- runTcM (tcCompUnit (CompUnit imps decls)) (initTcEnv options)
-          case r of
-            Left err1 -> pure $ Left err1
-            Right (CompUnit _ ds, env) -> do
-              let ds1 = (ds ++ generated env)
-              pure (Right (CompUnit imps ds1, env))
+        Left err1 -> pure $ Left err1
+        Right (CompUnit _ ds, env) -> do
+          let ds1 = (ds ++ generated env)
+          pure (Right (CompUnit imps ds1, env))
 
 -- type inference for a compilation unit
 
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -759,6 +759,10 @@ verifySignatures instd@(Instance _ _ ps n ts t funs) =
 
 checkMemberType :: (Name, Qual Ty, Qual Ty) -> TcM ()
 checkMemberType (qn, qt@(ps :=> t), qt'@(ps' :=> t'))
+  -- whenever we have a closure, the infered type
+  -- will change. This fact causes an error when
+  -- the function has a signature, since the infered
+  -- type will not match the annotated type.
   | hasClosureType t = pure ()
   | otherwise
     = do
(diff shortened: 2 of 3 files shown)


src/Solcore/Frontend/TypeInference/TcContract.hs
```diff
@@ -29,14 +29,18 @@ import Solcore.Primitives.Primitives
...
-        Right ((CompUnit imps ds), env) -> do
-          let ds1 = (ds ++ generated env)
-          pure (Right (CompUnit imps ds1, env))
+        Right _ -> do
+          r <- runTcM (tcCompUnit (CompUnit imps decls)) (initTcEnv options)
+          case r of
+            Left err1 -> pure $ Left err1
+            Right (CompUnit _ ds, env) -> do
+              let ds1 = (ds ++ generated env)
+              pure (Right (CompUnit imps ds1, env))
```
mbenke: I think these two passes should be more explicit in SolcorePipeline - then we may consider putting some desugaring in between. Both contract field desugaring and dispatch generation can benefit from type information.

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -810,7 +836,10 @@ checkInstance idef@(Instance d vs ctx n ts t funs)
       -- kind check all types in instance head
       mapM_ kindCheck (t : ts) `wrapError` idef
       -- check if the class is defined
-      _ <- askClassInfo n `wrapError` idef
+      cinfo <- askClassInfo n `wrapError` idef
+      -- check if the instance arity is correct
+      unless (length ts == classArity cinfo) $
+        classArityError n cinfo idef
```
mbenke, line 850: This seems to overlap with #310

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1194,6 +1223,14 @@ rename t = let vs = bv t
...
 -- errors
 
+classArityError :: Pretty a => Name -> ClassInfo -> a -> TcM ()
+classArityError n cinfo v
+  = throwError $ unlines [ "Type class:" ++ pretty n
+                         , "has arity of:" ++ show (classArity cinfo)
+                         , "which does not match:"
+                         , pretty v
+                         ]
+
```
mbenke, line 1241: This seems to overlap with #310

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -737,10 +759,14 @@ verifySignatures instd@(Instance _ _ ps n ts t funs) =
 
 checkMemberType :: (Name, Qual Ty, Qual Ty) -> TcM ()
 checkMemberType (qn, qt@(ps :=> t), qt'@(ps' :=> t'))
-  = do
-      _ <- tcmMatch t t' `catchError` (\ _ -> invalidMemberType qn t t')
-      pure ()
+  | hasClosureType t = pure ()
```
mbenke, line 766: If I understand it correctly, this can be a noop because of the first phase, right? An explanatory comment would be helpful here.

rodrigogribeiro: Added a comment explaining why this should be a no op.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: Rebased on top of the current main.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: Rebased on top of the current main.

src/Solcore/Frontend/TypeInference/TcContract.hs
rodrigogribeiro: Moved to SolcorePipeline.

Issue #295, opened 2025-12-24
Type checker doesn't type check for contract field access


Y-Nak:
It seems the compiler doesn't type-check the contract field access when the field access is the lhs of the assignment.

## PoC
```
import std;

contract PoC {
    field : word;

    function set_x(b: bool) -> bool {
        field = b;   // BUG: `word` shouldn't be unified with `bool`.
        return b;
    }

    function init(foo: bool) -> () {
       field = 2;
    }
}
```

rodrigogribeiro:
Fixed by #317.

PR #317, opened 2026-02-12
Some fixes on name resolution for fields.

### LINKED ISSUE #296: using `x` as a field name causes an compilation error
When `x` is used in the contract field, the compiler emits an error.
## PoC
Code:
```
import std;

contract PoC {
    x : word;
}
```
Error:
```
tcExp not implemented for: this.x
FieldAccess Nothing x
 - in:Typedef.rep(this.x)
 - in:function toWord (x : ?$5999) -> word {
   return Typedef.rep(this.x);
}
 - in:?$5999 : Add, ?$5999 : Sub, ?$5999 : Bounded, ?$5999 : Eq, ?$5999 : Ord, ?$5999 : Typedef (word) => default instance ?$5999 : Num {
   function maxVal () -> ?$5999 {
      return Bounded.maxVal();
   }
   function toWord (x : ?$5999) -> word {
      return Typedef.rep(this.x);
   }
   function fromWord (x : word) -> ?$5999 {
      return Typedef.abs(this.x);
   }
   function add (x : ?$5999, y : ?$5999) -> ?$5999 {
      return Add.add(this.x, y);
   }
   function sub (x : ?$5999, y : ?$5999) -> ?$5999 {
      return Sub.sub(this.x, y);
   }
   function gt (x : ?$5999, y : ?$5999) -> bool {
      return Ord.gt(this.x, y);
   }
}
```

If `x` is renamed to `x1`, for example, the compilation succeeds.

rodrigogribeiro:
This PR fixes #296 and #295.

### DIFF
--- a/src/Solcore/Frontend/Syntax/NameResolution.hs
+++ b/src/Solcore/Frontend/Syntax/NameResolution.hs
@@ -1,5 +1,6 @@
 module Solcore.Frontend.Syntax.NameResolution where
 
+import Common.Pretty
 import Control.Applicative
 import Control.Monad
 import Control.Monad.Except
@@ -7,7 +8,7 @@ import Control.Monad.State
 import Data.List ((\\))
 import Data.Map (Map)
 import Data.Map qualified as Map
-import Solcore.Frontend.Pretty.SolcorePretty
+import Solcore.Frontend.Pretty.TreePretty
 import Solcore.Frontend.Syntax.Contract hiding (contracts, decls)
 import Solcore.Frontend.Syntax.Name
 import Solcore.Frontend.Syntax.Stmt
@@ -48,29 +49,29 @@ instance (Resolve a) => Resolve (Maybe a) where
 instance Resolve S.TopDecl where
   type Result S.TopDecl = TopDecl Name
 
-  resolve (S.TContr c) =
-    TContr <$> withLocalCtx (resolve c)
-  resolve (S.TFunDef fd) =
-    TFunDef <$> withLocalCtx (resolve fd)
-  resolve (S.TClassDef c) =
-    TClassDef <$> withLocalCtx (resolve c)
-  resolve (S.TInstDef d) =
-    TInstDef <$> withLocalCtx (resolve d)
-  resolve (S.TDataDef dt) =
-    TDataDef <$> withLocalCtx (resolve dt)
-  resolve (S.TSym ts) =
-    TSym <$> withLocalCtx (resolve ts)
-  resolve (S.TPragmaDecl p) = TPragmaDecl <$> resolve p
+  resolve t@(S.TContr c) =
+    TContr <$> withLocalCtx (resolve c) `wrapError` t
+  resolve t@(S.TFunDef fd) =
+    TFunDef <$> withLocalCtx (resolve fd) `wrapError` t
+  resolve t@(S.TClassDef c) =
+    TClassDef <$> withLocalCtx (resolve c) `wrapError` t
+  resolve t@(S.TInstDef d) =
+    TInstDef <$> withLocalCtx (resolve d) `wrapError` t
+  resolve t@(S.TDataDef dt) =
+    TDataDef <$> withLocalCtx (resolve dt) `wrapError` t
+  resolve t@(S.TSym ts) =
+    TSym <$> withLocalCtx (resolve ts) `wrapError` t
+  resolve t@(S.TPragmaDecl p) = TPragmaDecl <$> resolve p `wrapError` t
 
 instance Resolve S.Contract where
   type Result S.Contract = Contract Name
 
-  resolve (S.Contract n vs decls) =
+  resolve c@(S.Contract n vs decls) =
     do
       let ns = map tyconName vs
       mapM_ addTyVar ns
       mapM_ addContractDecl decls
-      Contract n (map TVar ns) <$> resolve decls
+      Contract n (map TVar ns) <$> resolve decls `wrapError` c
 
 addContractDecl :: S.ContractDecl -> ResolveM ()
 addContractDecl (S.CDataDecl (S.DataTy n _ cons)) =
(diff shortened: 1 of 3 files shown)


src/Solcore/Pipeline/SolcorePipeline.hs
```diff
@@ -66,14 +66,12 @@ compile opts = runExceptT $ do
...
 
   -- Name resolution
-  (resolved, nameEnv) <- ExceptT $ buildAST' parsed
+  resolved <- ExceptT $ nameResolution parsed
 
   liftIO $ when (verbose || optDumpAST opts) $ do
     putStrLn "> AST after name resolution"
     putStrLn $ pretty resolved
 
-  liftIO $ when (optDumpEnv opts) $ pPrint nameEnv
```
Y-Nak, line 75: Now, `optDumpEnv` is not used. Probably, it'd be better to remove the flag from the CLI option.

src/Solcore/Frontend/Syntax/NameResolution.hs
```diff
@@ -0,0 +1,702 @@
...
+      e1' <- resolve e1
+      e2' <- resolve e2
+      let fun = QualName (Name "Div") "div"
+      pure $ Call Nothing fun [e1', e2']
+  resolve (S.ExpModulo e1 e2) =
+    do
+      e1' <- resolve e1
+      e2' <- resolve e2
+      let fun = QualName (Name "Mod") "mod"
+      pure $ Call Nothing fun [e1', e2']
```
Y-Nak, line 393: Currently, there is no instance of `Mod`. So, probably it'd be better to add instance `word: Mod` at least in this PR.
code: 
```
import std;
function foo(x: word, y: word) -> word {
    return x % y;
}
```

result:
```
Cannot entail:
word : Mod
using defined instances:


 - in:function foo (x : word, y : word) -> word {
   return Mod.mod(x, y);
}
```

src/Solcore/Frontend/Syntax/NameResolution.hs
```diff
@@ -0,0 +1,702 @@
...
+        (_, Just TParameter) -> pure (Var n)
+        -- field access
+        (Nothing, Just TField) ->
+          pure (FieldAccess Nothing n)
+        -- function reference
+        (_, Just TFunction) -> do
+          dt1 <- gets (Map.lookup n . fieldEnv)
+          case dt1 of
+            Just TField -> pure (FieldAccess Nothing n)
+            _ -> pure (Var n)
```
mbenke, line 315: Should fields have precedence over functions? Is there a way to call the function if there is a field with same name?

We should discuss it and maybe consider something like `this.field`

src/Solcore/Frontend/Syntax/NameResolution.hs
```diff
@@ -0,0 +1,702 @@
...
+
+-- definition of an environment
+
+data DeclType
+  = TContract
+  | TFunction
+  | TDataCon
+  | TLocalVar
+  | TParameter
+  | TPattern
```
mbenke, line 506: What is this for? Pattern synonyms?

std/std.solc
```diff
@@ -156,7 +156,6 @@ forall a. a:Ord =>
 function gt(x:a, y:a) -> bool {
   return Ord.gt(x,y);
 }
-
```
mbenke, line 159: why remove this line?

src/Solcore/Frontend/Syntax/NameResolution.hs
```diff
@@ -0,0 +1,702 @@
...
+        fdt = Map.lookup n (fieldEnv env)
+    pure (ldt <|> gdt <|> cdt <|> fdt)
+
+wrapError :: (Pretty b) => ResolveM a -> b -> ResolveM a
+wrapError m e =
+  catchError m handler
+  where
+    handler msg = throwError (decorate msg)
+    decorate msg = msg ++ "\n - in:" ++ pretty e
+
```
mbenke, line 638: This seems unused, do you plan to use it?

```suggestion
```

mbenke (commented):
ElabTree should probably be removed from sol-core.cabal

src/Solcore/Frontend/Syntax/NameResolution.hs
rodrigogribeiro: This ambiguity is what I said last week about using something like quoted names for functions or even use `this`. I must confess that I would prefer to use quotes... :)

src/Solcore/Frontend/Syntax/NameResolution.hs
rodrigogribeiro: This is to identify when a name is introduced by a pattern matching.

src/Solcore/Frontend/Syntax/NameResolution.hs
rodrigogribeiro: Currently, it is not used, but the error messages are crap. I intend to add `wrapError` to give context to error messages.

src/Solcore/Frontend/Syntax/NameResolution.hs
rodrigogribeiro: Fixed by adding a instance for `word` in `std.solc`. Added the test suggested by @Y-Nak in the test suite.

std/std.solc
rodrigogribeiro: Added this empty line again and an additional instance for `word:Mod`.

src/Solcore/Pipeline/SolcorePipeline.hs
rodrigogribeiro: Removed from CLI options.

Y-Nak (approved):
LGTM!

src/Solcore/Frontend/Syntax/NameResolution.hs
rodrigogribeiro: Added better error messages to name resolution.

PR #318, opened 2026-02-13
Partial evaluation for comptime word, boolean  and string values

mbenke:
Basic operations on words, boools and strings are evaluated at comptime:

**Primitives** (when all arguments are compile-time literals):

| Function | Operands | Result |
|---|---|---|
| `addWord` | `word, word` | `word` (integer addition) |
| `subWord` | `word, word` | `word` (integer subtraction) |
| `gtWord` | `word, word` | `bool` (greater-than comparison) |
| `eqWord` | `word, word` | `bool` (equality comparison) |
| `concatLit` | `string, string` | `string` (concatenation) |
| `strlenLit` | `string` | `word` (UTF-8 byte length) |
| `keccakLit` | `string` | `word` (keccak256 hash as 256-bit big-endian integer) |

**Structural:**

- **Variable propagation** — known values tracked in VEnv, substituted at use sites
- **Function inlining** — pure functions inlined when the evaluator has fuel (fuel limits call depth, not total inlines)
- **Match branch selection** — `MastMatch` selects the matching branch when the scrutinee is a known constructor
- **Dead expression elimination** — `MastStmtExp` dropped when the value is compile-time known (no side effects)

**Tests** in `test/examples/comptime`

**Not yet optimized:**

- `MastCond` — condition always evaluated, both branches kept even when the condition is known

The key downstream use case is **method selector computation**: `concatLit` chains build the ABI signature string, `keccakLit` hashes it, and `addWord`/`subWord` handle any arithmetic — all folded to a single integer constant.

### DIFF
--- a/src/Solcore/Backend/MastEval.hs
+++ b/src/Solcore/Backend/MastEval.hs
@@ -48,14 +48,22 @@ defaultFuel = 100
 -- Evaluation monad
 -----------------------------------------------------------------------
 
--- Reader for constant function table, State for fuel budget
-type EvalM = ReaderT FunTable (State Fuel)
+data EvalEnv = EvalEnv
+  { envFunTable :: FunTable,
+    envPureFuns :: Set.Set Name
+  }
 
-runEvalM :: FunTable -> Fuel -> EvalM a -> (a, Fuel)
-runEvalM ft fuel m = runState (runReaderT m ft) fuel
+-- Reader for constant environment, State for fuel budget
+type EvalM = ReaderT EvalEnv (State Fuel)
+
+runEvalM :: EvalEnv -> Fuel -> EvalM a -> (a, Fuel)
+runEvalM env fuel m = runState (runReaderT m env) fuel
 
 askFunTable :: EvalM FunTable
-askFunTable = ask
+askFunTable = asks envFunTable
+
+askPureFuns :: EvalM (Set.Set Name)
+askPureFuns = asks envPureFuns
 
 getFuel :: EvalM Fuel
 getFuel = lift get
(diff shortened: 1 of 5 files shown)


rodrigogribeiro (approved):
Great work!

src/Solcore/Backend/MastEval.hs
```diff
@@ -0,0 +1,498 @@
...
+          _ -> Map.delete i env
+    evalFunBody env' rest
+  MastAssign i e -> do
+    e' <- evalExp env e
+    let env' =
+          if isKnownValue e'
+            then Map.insert i e' env
+            else Map.delete i env
+    evalFunBody env' rest
+  MastStmtExp _ -> evalFunBody env rest
```
Y-Nak, line 330: This seems problematic because this code ignores all side-effectual statements. 
The problem is
1. `isInlinable` function marks any `MastStmtExp` as inlinable
2. Because of 1., a function that has side-effectful statements is inlined, and then this code just ignores these side-effectful statements.
3. If the returned expression can be evaluated, whole side-effectful statements in the function are just removed.

e.g.,
code:
```
import std;

contract Foo {
  function my_revert() -> word {
    revert("regression");
    return 0;
  }

  function main() -> word {
    my_revert();
    return 1;
  }
}
```
`output1.hull`:
```
object Foo {
  code {
    function main () -> word {
      return 1
    }
  }
}
```

mbenke: Good catch! I have added a purity check - for now it treats revert and all assembly blocks as impure, eventually we'll need to add a Yul interpreter, but I think not until we have resolved string handling (and memory in general) in comptime assembly.

PR #319, opened 2026-02-13
Dispatch: replace runtime selector computation with compile-time

mbenke:
Method selectors (4-byte function identifiers used for EVM dispatch) were computed at runtime: `ContractDispatch.hs` generated `ABIString` instances that emitted Yul assembly to build ABI signature strings in memory byte-by-byte, then hashed them with `keccak256`. This was complex and wasteful since signature strings are fully known at compile time.

This PR replaces the runtime approach with compile-time computation using the `SigString` typeclass. `ContractDispatch.hs` now generates `SigString` instances that return the method name as a string literal. `dispatch.solc` builds the full signature via string concatenation (`concatLit`) and hashes it with `keccakLit` - the partial evaluator folds the entire chain to an integer constant, so no string operations or hashing remain at runtime.

### Changes

- **`ContractDispatch.hs`**: `mkNameInst` generates `SigString` instances (single `sigStr` method returning a `StrLit`) instead of `ABIString` instances (Yul byte-manipulation assembly). Removed `generateStoreBytes`, `storeChunk`, `chunk32`, `bsToInteger` and their imports.
- **`std/dispatch.solc`**: Replaced with `ABIString` the `SigString`-based dispatch
- **`std/rtdispatch.solc`**: Old runtime `ABIString` dispatch preserved for reference.
- **`std/std.solc`**: Added `string:Add` instance delegating to `concatLit`.

### Testing

`./run_contests.sh` runs tests from `test/examples/dispatch` ensuring dispatch works correctly after changes (including `miniERC20.solc`, which is the main litmus test).

### Potential TODO

Replace chained `if`s by a big match - this would require allowing comptime expressions in patterns - TBD

### DIFF
--- a/src/Solcore/Desugarer/ContractDispatch.hs
+++ b/src/Solcore/Desugarer/ContractDispatch.hs
@@ -207,4 +207,3 @@ nameTypeName cname fname = Name ("DispatchNameTy_" <> nm cname <> "_" <> nm fnam
   where
     nm (Name s) = s
     nm (QualName _ s) = s
-


test/examples/dispatch/miniERC20.json
rodrigogribeiro, line 1: Is there a reason for this file be in this commit or is it a temporary generated during tests?

test/examples/dispatch/basic.json
rodrigogribeiro, line 1: Is this file necessary in this commit or is it a temporary?

src/Solcore/Pipeline/SolcorePipeline.hs
```diff
@@ -183,10 +187,28 @@ compile opts = runExceptT $ do
...
+          putStrLn "!! Warning: partial evaluation ran out of fuel (use --pe-fuel N to increase)"
+
+      liftIO $ when (optDumpSpec opts) $ do
+        putStrLn "> After partial evaluation:"
+        putStrLn (pretty evaluated)
+
+      -- Dead code elimination: remove functions unreachable from 'start'/'main'
+      let optimized = eliminateDeadCode evaluated
+
+      liftIO $ when (optDumpSpec opts) $ do
```
rodrigogribeiro, line 204: Is this option enabled by default? I'd argue that it would be useful to reduce the noise of the compiler output when using the standard library.

test/examples/dispatch/basic.json
mbenke: `test/examples/dispatch/*.json` files are used by `run_contests.sh` which is part of #314 - I forgot this has not been merged yet

src/Solcore/Pipeline/SolcorePipeline.hs
mbenke: No, it is enabled by the `--dump-spec` flag (so this is not printed even with `-v`)

test/examples/dispatch/basic.json
rodrigogribeiro: No worries... after checking the other PR I noticed that. Sorry for the noise.

PR #320, opened 2026-02-24
New proxy syntax

rodrigogribeiro:
This PR adds a simple syntax sugar for proxies.

* At type level, @type denotes Proxy(type).
* At expression level, @type denotes Proxy : Proxy(type).


Y-Nak (approved):
LGTM!

PR #323, opened 2026-03-03
Contest2

Y-Nak:
This PR is a successor to #276.

The main changes from #276 are 
1. Contract test is now integrated into`cabal`
2. By default, tests fail if testrunner/evmone are missing (skip only with `SOLCORE_CONTRACT_TESTS_ALLOW_SKIP=1`)
3. `intx` and `blst` are pinned in `nix/evmone-pins.nix` and checked against evmone’s own CMake pins at build time

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -29,4 +29,24 @@ jobs:
     - uses: cachix/install-nix-action@v31
       with:
         github_access_token: ${{ secrets.GITHUB_TOKEN }}
-    - run: nix build .#checks.x86_64-linux.ormolu
+    - run: nix build -L .#checks.x86_64-linux.ormolu
+
+  tests-no-warnings:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: cachix/install-nix-action@v31
+      with:
+        github_access_token: ${{ secrets.GITHUB_TOKEN }}
+    - run: |
+        set -euo pipefail
+        nix develop --command bash -lc '
+          set -euo pipefail
+          cabal update
+          cabal clean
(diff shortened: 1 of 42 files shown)


run_contests.sh
```diff
@@ -1,4 +1,10 @@
...
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "$root_dir"
+
+bash ./contest.sh test/examples/dispatch/basic.json
+bash ./contest.sh test/examples/dispatch/neg.json
+# bash ./contest.sh test/examples/dispatch/miniERC20.json
```
mbenke: Why is miniERC20 commented out?

testsol.sh
```diff
@@ -10,21 +10,21 @@ function runsol() {
     file=$1
     echo $file
     shift
-    rm -f -v output1.hull Output.sol
+    rm -f -v output1.core Output.sol
```
mbenke: That's probably my fault, but this seems a regression - better just use runsol.sh from the `main` branch

Y-Nak: Fixed.

run_contests.sh
Y-Nak: oops. Fixed.

mbenke:
Sorry, it partly my fault - it seems that the directory wasn't as clean as I thought - `git worktree` copied tainted submodules there. But the point about thw error still stands:

```
    if [[ ! -f "$root_dir/deps/nlohmann_json/CMakeLists.txt" ]] && command -v git >/dev/null 2>&1; then
        git -C "$root_dir" submodule update --init --depth 1 deps/nlohmann_json >/dev/null 2>&1 || true
    fi
```

causes errors to be silently ignored here and explode in cmake.

PR #325, opened 2026-03-11
add evmone to devShells buildInputs and shellHook

mbenke:


flake.nix
```diff
@@ -141,10 +141,14 @@
             pkgs.go-ethereum
             pkgs.jq
             pkgs.solc
+            evmone-lib
             (hspkgs.hevm.overrideAttrs (old: { patches = []; }))
             texlive
             (pkgs.callPackage ./nix/goevmlab.nix { src = inputs.goevmlab; })
           ];
+          shellHook = ''
+            export evmone="${evmone-lib}/lib/libevmone.so"
```
d-xo: this feels a little dirty, where is this actually required?

mbenke: Indeed, I am groping in the dark mostly, but `evmone` var is needed in the contract test, which is the whole point; without it we get:

```
$ ./scripts/cabal-contract-tests.sh 
[sol-core-contract-test] libevmone not found. Set evmone=/absolute/path/to/libevmone.{so,dylib}.
```

can you suggest a better solution? Otherwise maybe @Y-Nak will solve it more systematically.

mbenke: Is it less dirty now, without the `shellHook` ?

PR #328, opened 2026-03-11
Fixes on type inference involving phantom types.

rodrigogribeiro:
This PR solves an ambiguity problem which could arise during type checking expression involving phantom types. 
Such expressions could have type variables which will not be further instantiated by type inference and this could cause the specialiser to crash. A example of this problems is the following program

```
data Foo(a) = Foo(word);

function read(x) {
  let res : word;
  match (x) {
  | Foo(w) =>
    assembly {
      res := w
    }
  }
  return res;
}

contract Bla {

  function main () -> word {
    return read(Foo(42));
  }
}
```
Expression `Foo(42)` has type `Foo(a)`, for some `a`, and this variable cannot be further instantiated by type inference. Solution for it to type check correctly is to add a type annotation which fixes `a` to some concrete type.


src/Solcore/Frontend/TypeInference/InvokeGen.hs
```diff
@@ -95,16 +95,9 @@ freshPatArg ty@(TyCon pn _) =
...
-                pure (PVar n, Var n)
-              else do
-                (ps', es) <- unzip <$> mapM freshPatArg args
-                pure (PCon cn ps', Con cn es)
+          [_cn] -> do
+            n <- freshName
+            pure (PVar n, Var n)
           _ -> do
             n <- freshName
             pure (PVar n, Var n)
```
mbenke: Isn't the whole `case` pointless now? it yields `pure (PVar n, Var n)` in both cases

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -626,6 +630,39 @@ ambiguous :: Scheme -> Bool
 ambiguous (Forall _ (ps :=> t)) =
   not $ null $ bv ps \\ bv (closure ps (bv t))
 
+-- Check for phantom meta variables: meta vars appearing in constructor
+-- application result types in the body but not in the function's inferred
+-- type or environment. Such variables arise when a constructor has phantom
+-- type parameters (type parameters that do not appear in any constructor
+-- field). They cannot be determined from context and indicate a type error.
+checkPhantomMetaVars :: Name -> Body Id -> [Pred] -> Ty -> TcM ()
```
mbenke: The check only fires on Con nodes in the typed body. But what about constructor patterns on the LHS of matches? Here's a case where the phantom meta var enters through a function call return type, not a Con expression:     

``` solidity                                  
  data Foo(a) = Foo(word);                                                                                       
                                                                                                                 
  function wrap(x) {                                                                                             
    return Foo(x);                                                                                               
  }                         

  function unwrap() -> word {
    match(wrap(42)) {
    | Foo(w) => return w;
    }
  }

  contract C {
    function main() -> word {
      return unwrap();
    }
  }
```

  In `unwrap`, the phantom a from `Foo(a)` is introduced by instantiating wrap's return type, not by a Con node.
`conResultMetaVars` traverses the body looking for `Con (Id _ ty)` args expressions - it won't find any in unwrap.
  Meanwhile `wrap` itself passes the check because the phantom var appears in its return type. But later, we crash in the specialiser:

```
PANIC: spec(wrap<word -> Foo($77)>(42)): free type vars in word -> Foo($77): [TVar {var = $77}] @ subst={a |-> $77}
```

src/Solcore/Frontend/TypeInference/InvokeGen.hs
rodrigogribeiro: Fixed.

src/Solcore/Frontend/TypeInference/TcStmt.hs
rodrigogribeiro: Good catch! I've fixed this and added you example to the test suite. Thanks!

src/Solcore/Frontend/TypeInference/InvokeGen.hs
mbenke: Am I missing something, or can now the whole `freshPatArg` be reduced to

``` haskell
freshPatArg _ =
  do
    n <- freshName
    pure (PVar n, Var n)
```
?

src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -626,6 +630,72 @@ ambiguous :: Scheme -> Bool
...
+checkPhantomMetaVars :: Bool -> Name -> Body Id -> [Pred] -> Ty -> TcM ()
+checkPhantomMetaVars checkReturn n body rs ty = do
+  envVars <- getEnvMetaVars
+  bodySubst <- withCurrentSubst body
+  tySubst <- withCurrentSubst (rs, ty)
+  let (rsApplied, tyApplied) = tySubst
+      legitimateMVs = mv tySubst ++ envVars
+      conMVs = conResultMetaVars bodySubst
+      -- Case 1: constructor-result MVs entirely absent from the function's type.
+      phantomMVs = conMVs \\ legitimateMVs
```
Y-Nak, line 654: This could lead to too strong an approximation for ambiguous phantom uses.
E.g., the current implementation rejects the code below.
```
data Foo(a) = Foo(word);

contract C {
  function main() -> word {
    match (Foo(1)) {
    | Foo(w) => return w;
    }
  }
}
```

I can't say this is a bug since this depends on a language spec(Rust does reject this case, but Haskell allows), but at least we need to decide a direction.

Y-Nak: Sorry, my comment was not right. I missed your comment, and completely misunderstood the PR. Please ignore this.

src/Solcore/Frontend/TypeInference/InvokeGen.hs
rodrigogribeiro: Fixed.

mbenke: There's still the TyCon case left. Does `maybeExpandTypeSynonym` have side effects we want to keep? If so, it should be documented. Otherwise I guess we can remove this case as well.

rodrigogribeiro: The type constructor might be a synonym, that needs to be expanded. Note that synonym expansion is done during type inference, so locations where types can appear in the program need to call expansion.

mbenke: But even then, with current changes `freshPatArg` seems to return ` (PVar n, Var n)` albeit in a convoluted way. In what case can/should it  something else?

mbenke: IOW, all test pass if `freshPatArg` is changed to

``` hakell
freshPatArg _ =
  do
    n <- freshName
    pure (PVar n, Var n)
```

can you suggest a test that would justify the current code?

rodrigogribeiro: You are right. This simpler definition works fine. However, the following program is causing an error:

```
type W = word;

function f(x:W) -> W { x }

contract C {

  function main () -> word {
    return f(42);
  }
}
```

Since after type inference, it's elaboration is (omitting invokable instances):

```
type W = word
data CCxt = CCxt  ;
function f (x<W> : W) -> W {
   return x<W>;
}
contract C {
   function main () -> word {
      return f<word -> word>(42);
   }
}
```
which does not expand the synonyms in the function definition for `f`, which causes the specializer to panic. 

@Y-Nak : Could you have a look on why the synonym `W` is not being expanded on the definition of `f`?

mbenke: But this happens also on main, as well as in the current version of this PR.
In fact the specialiser was not written with type synonyms in mind, but could probably be extended for them as well.

mbenke: Created a separate issue for this: #338

rodrigogribeiro: Fixed in #339.

PR #329, opened 2026-03-11
Adding a check to avoid recursive type definitions

rodrigogribeiro:
* This PR adds a check which do not allow the declaration of recursive types (direct or indirect).
* Add new tests and changes some tests to failing due to the use of recursive types.


test/Cases.hs
```diff
@@ -102,7 +102,7 @@ pragmas =
   testGroup
     "Files for pragmas cases"
     [ runTestExpectingFailure "bound.solc" pragmaFolder,
-      runTestForFile "coverage.solc" pragmaFolder,
+      runTestExpectingFailure "coverage.solc" pragmaFolder,
```
mbenke: As is, this test is useless,consider switching to memory-based lists:

```
data List(a) = Nil | Cons(a,memory(List(a)));
```

test/Cases.hs
```diff
@@ -112,7 +112,7 @@ cases :: TestTree
 cases =
   testGroup
     "Files for folder cases"
-    [ runTestForFile "Ackermann.solc" caseFolder,
+    [ runTestExpectingFailure "Ackermann.solc" caseFolder,
```
mbenke: Consider switching to memory Nat:

```
data memory(a)  = memory(word);
data Nat = Zero | Succ(memory(Nat)) ;
```

test/Cases.hs
```diff
@@ -152,12 +152,12 @@ cases =
       runTestExpectingFailure "duplicated-type-name.solc" caseFolder,
       runTestForFile "DuplicateFun.solc" caseFolder,
       runTestExpectingFailure "DupFun.solc" caseFolder,
-      runTestForFile "EitherModule.solc" caseFolder,
+      runTestExpectingFailure "EitherModule.solc" caseFolder,
```
mbenke, line 267: consider switching to memory-based lists; this would also enable us to see if memory types are viable

mbenke (commented):
I wonder whether just switching tests to 'expect failure' is the right way to go. 

Also, we should think whether we want to allow recursive types in comptime.

test/Cases.hs
rodrigogribeiro: Fixed.

test/Cases.hs
rodrigogribeiro: Fixed.

test/Cases.hs
rodrigogribeiro: Fixed.

src/Solcore/Frontend/TypeInference/TcContract.hs
```diff
@@ -101,6 +108,76 @@ recursiveSynonymError cyclePath =
...
+
+buildPhantomMap :: [DataTy] -> Map.Map Name (Set Int)
+buildPhantomMap = Map.fromList . map phantomEntry
+  where
+    phantomEntry (DataTy n params ctors) =
+      let allFieldTys = concatMap constrTy ctors
+          usedVarNames = concatMap tyVarNames allFieldTys
+          isPhantomParam p = tyvarName p `notElem` usedVarNames
+          phantomIdxs = Set.fromList [i | (i, p) <- zip [0 ..] params, isPhantomParam p]
+       in (n, phantomIdxs)
```
Y-Nak, line 227: This function only recognizes direct uses. A wrapper over phantom type is regarded as having a non-phantom param, as a result, this can lead to an over-approximation on cycle detection.
e.g.,
```
data Box(a) = Box(memory(a));
data Rec = Rec(Box(Rec));
```
A compiler rejects the code, because `Rec` is recognized as a recursive type definition, but actually, this is not a recursive definition.

Y-Nak (changes requested):

src/Solcore/Frontend/TypeInference/TcContract.hs
mbenke: Perhaps instead of checking for recursion, we should try to find an upper bound on size of elements of the type and reject if it is not finite or too big for the stack (since this is actually the issue).

This may be too low-level for a type-checker though.

rodrigogribeiro: I've changed the algorithm in order to build a phantom-parameter map using fixpoint iteration so that
transitively-phantom positions are discovered.

src/Solcore/Frontend/TypeInference/TcContract.hs
```diff
@@ -101,6 +108,102 @@ recursiveSynonymError cyclePath =
...
+        . concatMap (\(Constr _ tys) -> concatMap (nonPhantomTyNames phantomMap) tys)
+        $ ctors
+
+checkRecursiveTypes :: [DataTy] -> TcM ()
+checkRecursiveTypes dts =
+  case cyclicSccs of
+    [] -> pure ()
+    (c : _) -> recursiveTypeError (NAG.vertexList1 c)
+  where
+    userTypes = Set.fromList (map dataName dts)
```
Y-Nak, line 258: This could be problematic when ADTs with the same name are defined in different contracts, since they will get merged. e.g.,
```
contract A {
  data T = MkT(U);
  data U = MkU(word);
}

contract B {
  data T = MkT2(word);
  data U = MkU2(T);
}
```

src/Solcore/Frontend/TypeInference/TcContract.hs
```diff
@@ -101,6 +108,102 @@ recursiveSynonymError cyclePath =
...
+  overlay isolated edged
+  where
+    phantomMap = buildPhantomMap dts
+    isolated = vertices (Set.toList userTypes)
+    edged = stars [(dataName dt, deps dt) | dt <- dts]
+    deps (DataTy _ _ ctors) =
+      nub
+        . filter (`Set.member` userTypes)
+        . concatMap (\(Constr _ tys) -> concatMap (nonPhantomTyNames phantomMap) tys)
+        $ ctors
```
Y-Nak, line 250: This doesn't expand type-synonyms and thus can miss a type recursion.
e.g.,
```
type Ref = T;
data T = Mk(Ref);
```

Y-Nak (changes requested):

src/Solcore/Frontend/TypeInference/TcContract.hs
rodrigogribeiro: Fixed and added as a test case.

src/Solcore/Frontend/TypeInference/TcContract.hs
rodrigogribeiro: Fixed and added as a test case.

PR #345, opened 2026-03-31
Fixes on type synonym expansion

rodrigogribeiro:
* This PR changes the implementation of type synonym expansion to use SYB and expand every synonym before start type inference.
* Adds a test case suggested by mbenke.
*  Fixes #344.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcContract.hs
+++ b/src/Solcore/Frontend/TypeInference/TcContract.hs
@@ -41,16 +41,25 @@ buildSynTable :: [TySym] -> SynTable
 buildSynTable syns =
   Map.fromList [(symName s, SynInfo (length (symVars s)) (symVars s) (symType s)) | s <- syns]
 
--- Pure recursive expansion of a single Ty using a SynTable
-expandTyPure :: SynTable -> Ty -> Ty
-expandTyPure st (TyCon n ts) =
-  let ts' = map (expandTyPure st) ts
-   in case Map.lookup n st of
-        Just (SynInfo _ params body)
-          | length params == length ts' -> expandTyPure st (insts (zip params ts') body)
-        _ -> TyCon n ts'
-expandTyPure st (t1 :-> t2) = expandTyPure st t1 :-> expandTyPure st t2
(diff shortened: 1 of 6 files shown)


mbenke (approved):
Looks good, I have one small gripe though: arity mismatches now lead to undefined type error, e.g.

```
Undefined type: F 
 
# tcFunDef function main () -> F(word, word) {
   return (42, 0);
}
## predicates in signature:()

 - in:F(word, word)
ben@trawa:~/work/review$ cat test/examples/cases/synonym-arity-mismatch.solc
type F(a) = pair(a, word);

function main() -> F(word, word) {
    return pair(42, 0);
}
```

whereas on main we get better message:

```
Type synonym F expects 1 arguments
but was given 2
```

src/Solcore/Frontend/TypeInference/TcContract.hs
```diff
@@ -36,20 +36,37 @@ typeInfer opts (CompUnit imps topDecls) =
...
+  Map.fromList [(symName s, SynInfo (length (symVars s)) (symVars s) (symType s)) | s <- syns]
+
+-- Pure recursive expansion of a single Ty using a SynTable
+expandTyPure :: SynTable -> Ty -> Ty
+expandTyPure st (TyCon n ts) =
+  let ts' = map (expandTyPure st) ts
+   in case Map.lookup n st of
+        Just (SynInfo _ params body)
+          | length params == length ts' -> expandTyPure st (insts (zip params ts') body)
+        _ -> TyCon n ts'
```
mbenke: I think this causes problems on arity mismatch.

mbenke: To be clear: my gripe about error handling is not a blocker - the error  messages will need a revamp anyway and I see that fixing this locally may be a pain as `expandTyPure` is ...well, pure.

rodrigogribeiro: Fixed in the last commit. I've made the expansion algorithm monadic in order to deal better with error messages.

PR #347, opened 2026-04-06
Fix pattern matching

Y-Nak:
* Fix a crash when a variable-only match branch contains multiple statements
* Avoid duplicating scrutinee evaluation in variable match branches
* Limit `ormolu` check to tracked source directories

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -69,7 +69,7 @@
             src = gitignore ./.;
           } ''
             cd $src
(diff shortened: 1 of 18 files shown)


mbenke:
Maybe we should consider adding a block statement to the abstract syntax(es). I do not feel good about the `Match [] [([], body)]` hack.

rodrigogribeiro (approved):
Great! Thanks!

PR #348, opened 2026-04-07
Allow comptime expressions as match labels

mbenke:
Allow stuff like

```
  function classify(x : word) -> word {
    let two : comptime word = 2;
    match x {
      | comptime two + 0             => return 2;
      | comptime toWord(uint256(40)) => return 40;
      | other                        => return other;
    }
  }
```

The idea is to use it for dispatch instead of chained if statements.


mbenke:
Included in #354

PR #349, opened 2026-04-08
EmitHull: better handling of nested matches

mbenke:
An example for the blog post revealed a long standing TODO in code emission for nested matches:

```
data AuctionState =
    NotStarted(word)
  | Active(word, address)
  | Ended(word, address)
  | Cancelled(word, address);

data Phase = Early | Late;

function discount(state : AuctionState, phase : Phase) -> word {
    match state, phase {
    | Active(bid, _), Early => return bid / 10;
    | Active(bid, _), Late  => return bid / 20;
    | _, _                  => return 0;
    }
}

contract Discount {
  function main() -> word {
    discount(Active(420,address(0)), Early)
  }
}
```
This PR introduces proper fresh name handling for match branch targets, rather then having them all called `$alt`

### DIFF
--- a/src/Solcore/Backend/EmitHull.hs
+++ b/src/Solcore/Backend/EmitHull.hs
@@ -41,7 +41,8 @@ data EcState = EcState
     ecNest :: Int,
     ecDebug :: Bool,
     ecContext :: [String],
-    ecDeployer :: Maybe Hull.Body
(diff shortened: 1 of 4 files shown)


rodrigogribeiro (approved):
Looks good! Thanks!

PR #350, opened 2026-04-09
Add implicit contract constructor if no explicit one defined

mbenke:
In dispatch generation add a default empty constructor to the contract if it does not have one. 
Should also remove many '!! Warning: no resolution found for start'  encountered when running tests (most of which have no sonstructor, or no contract even).


Y-Nak (approved):
LGTM. 
One thing I realized is that if a contract has a `start` function, then a compilation fails; `Duplicated function definition:start`.
This is not directly related to this PR, but we might need to consider how to lower the constract constructor.

mbenke:
@Y-Nak thanks, good catch! This is a bigger change then it seemed at first glance though, so I'll do that in a separate PR.

PR #343, opened 2026-03-25
Add the comptime keyword (with no effect yet)

mbenke:
This PR introduces syntax - just syntax - for annotating parameters with `comptime` keyword, e.g.

``` solidity
contract ComptimeSyntax {

  function f(comptime x : word) -> comptime word {
    return x;
  }

  function g() -> word {
    let y : comptime word = f(42);
    return y;
  }

  function main() {
    return g();
  }
}
```

PR #351, opened 2026-04-09
Comptime annotation check

mbenke:
This PR introduces two passes checking `comptime` annotations on parameters, results and let-bound variables (introduced by #343).

The first pass is done on SAIL level, however for polymorphic and overloaded functions sometimes this is at least difficult to decide before specialisation, and  these cases are deferred to MAST-level checker.


mbenke:
Included in #354

PR #354, opened 2026-04-16
Comptime evaluation of some assembly blocks

mbenke:
This PR adds the ability to interpret assembly blocks when evaluating comptime expressions/

Blocks involving memory access are evaluated only when in “comptime mode” which is triggered by `let x : comptime ...`

Blocks containing only pure operations (arithmetic) are evaluated in “folding mode” (needed since even addition is defined via assembly).

Blocks containing runtime-only operations (such as `sload`) are not interpreted.


rodrigogribeiro (approved):
Great work!

PR #356, opened 2026-04-16
Add ownable contract example and integration test

czepluch:
Simple contract with owner field, getter, and guarded setter.
Tests cover deployment, reading owner, transferring ownership, and rejection of unauthorized ownership change.
Added test to `run_contests.sh`


d-xo (approved):
lgtm

PR #357, opened 2026-04-19
Documentation about Hull IR.

rodrigogribeiro:


doc/src/hull/README.md
```diff
@@ -1,256 +1,623 @@
 # Hull
 
-## Introduction
+Hull is an intermediate language used by the Core Solidity compiler between the
+typed frontend and the Yul backend. Every Core Solidity contract is lowered to
```
d-xo:
```suggestion
Hull is an intermediate language used by the Core Solidity compiler between SAIL and the Yul backend. Every SAIL program is lowered to
```

This is probably more accurate?

doc/src/hull/README.md
```diff
@@ -1,256 +1,623 @@
...
+Hull after type-checking, monomorphization, and match compilation. Hull is then
+translated to Yul for final code generation.
 
-The Hull language is an intermediate step before Yul generation. It is basically Yul with algebraic
-types (sums/products)
+Hull retains algebraic data types — products and sums — from the source
+language, but eliminates polymorphism, type classes, higher-order functions, and
+all surface syntax sugar. The result is a first-order, monomorphic language
+whose structure maps directly onto Yul constructs.
```
d-xo:
```suggestion
Hull retains algebraic data types — products and sums — from SAIL, but eliminates polymorphism, type classes, higher-order functions, and
all surface syntax sugar. The result is a first-order, monomorphic language
whose structure maps directly onto Yul constructs.
```

doc/src/hull/README.md
```diff
@@ -1,256 +1,623 @@
...
+
+It holds a value of either type `T1`, tagged as the left branch, or type `T2`,
+tagged as the right branch. The `+` operator is right-associative, so
+`A + B + C` is parsed as `A + (B + C)`.
+
+Sum values are constructed using injection expressions (`inl`, `inr`, `in(k)`)
+and deconstructed using `match` statements.
+
+An n-ary sum can also be written as `sum(T1, T2, ..., Tn)`, but the binary
+right-nested encoding is what the Yul code generator uses internally.
```
d-xo, line 84: curious if there is any reason to support both syntactic forms? or can we just drop the `sum(...)` version if we don't use it?

d-xo (approved):
very nice!!! just a few small comments where I think we could be a little more precise about the core solidity vs sail distinction...

doc/src/hull/README.md
rodrigogribeiro: I remember that @mbenke used n-ary sums to compress tags during Yul lowering.

mbenke: When I wrote this code, Daniel insisted on binary sums, so I have kept binary sums as a basic form and n-ary as an optimisation. We can probably change it, but it's orthogonal to the docs.

mbenke (approved):
Nicely done.

mbenke:
Will it be integrated with docs on https://argotorg.github.io/solcore ?

PR #361, opened 2026-04-29
Update README with Nix usage instructions

axic:
Added note about not running 'git submodule update --init' as Nix manages dependencies. Running it prior causes a build error.


mbenke:
Please merge the current main so that the checks pass.

mbenke (approved):
LGTM

PR #363, opened 2026-04-29
Fixes on yule generation, to avoid wrong renaming of variables.

rodrigogribeiro:
This PR fixes a problem in the yul code emission. Consider the following snippet: 

```
  function deposit(pubkey: memory(string), withdrawal_credentials: memory(string), signature: memory(string), deposit_data_root: uint256) -> () {
    let msg_value : uint256 = uint256(0);
    assembly {
        msg_value := callvalue()
    }
}
```

The emitted Yul eliminates the definition of `msg_value`:

```
function usr$deposit (pubkey, withdrawal_credentials, signature, deposit_data_root) {
        let _v382
        _v382 := 0
        msg_value := callvalue()
}
```

### DIFF
--- a/test/examples/cases/yul-deposit-example.solc
+++ b/test/examples/cases/yul-deposit-example.solc
@@ -1,0 +1,14 @@
+import std.{*};
+
+function deposit(pubkey: memory(string), withdrawal_credentials: memory(string), signature: memory(string), deposit_data_root: uint256) -> () {
+    let msg_value : uint256 = uint256(0);
+    assembly {
+        msg_value := callvalue()
+    }
+}
+
+contract Foo {
+   function main () -> () {
+      deposit(memory(0), memory(0), memory(0), uint256(2));
+   }
+}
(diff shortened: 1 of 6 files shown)


test/examples/cases/yul-deposit-example.solc
axic, line 1: I'd suggest removing the `deposit` part and just making it a generic `function f`. It has little do with the deposit contract and we may have many other examples/bugs coming from there.

mbenke:
Good catch, but we may need to handle also `YLet`, `YBlock`, `YSwitch` and `YFor` in `substAsmStmt` as well

mbenke:
Indeed, at least `YFor` and `YSwitch`:

```
import std.{*};

function yul_asm_for_body() -> () {
    let result : uint256 = uint256(0);
    assembly {
        for { let i := 0 } lt(i, 3) { i := add(i, 1) } {
            result := callvalue()
        }
    }
}

contract Foo {
    function main() -> () {
        yul_asm_for_body()
    }
}
```

```
import std.{*};

function yul_asm_switch_body() -> () {
    let result : uint256 = uint256(0);
    let flag : word = 1;
    assembly {
        switch flag
        case 0 { result := 0 }
        default { result := callvalue() }
    }
}

contract Foo {
    function main() -> () {
        yul_asm_switch_body()
    }
}
```

mbenke (approved):
Looks good now. I took the liberty of adding some tests.

PR #368, opened 2026-05-04
Implementing a type checker for Hull

rodrigogribeiro:
- This PR introduces a type checker for Hull into Yule. 
- Adds test cases for simple Hull programs that are accepted and rejected by the type checker.


src/Language/Hull/TcEnv.hs
```diff
@@ -0,0 +1,142 @@
...
+      ("tstore", HullFunSig w2 TUnit),
+      ("return", HullFunSig w2 TUnit),
+      ("revert", HullFunSig w2 TUnit),
+      ("calldatacopy", HullFunSig w3 TUnit),
+      ("codecopy", HullFunSig w3 TUnit),
+      ("returndatacopy", HullFunSig w3 TUnit),
+      ("mcopy", HullFunSig w3 TUnit),
+      ("extcodecopy", HullFunSig w3 TUnit),
+      ("create", HullFunSig w3 TUnit),
+      ("create2", HullFunSig w3 TUnit),
```
mbenke: See https://docs.soliditylang.org/en/latest/yul.html


```suggestion
      ("create", HullFunSig w3 TWord),
      ("create2", HullFunSig w4 TWord),
```

Unless I am mistaken, the signatures for these are wrong also in `Primitives.hs`

src/Language/Hull/TcEnv.hs
```diff
@@ -0,0 +1,142 @@
...
+
+emptyHullTcEnv :: HullTcEnv
+emptyHullTcEnv = HullTcEnv Map.empty hullBuiltins Nothing
+
+-- Number of word-sized slots a return type occupies.
+returnCount :: Type -> Int
+returnCount TUnit = 0
+returnCount TWord = 1
+returnCount (TPair a b) = returnCount a + returnCount b
+returnCount _ = 1
```
mbenke, line 35: This does not seem to handle `TNamed` correctly
```suggestion
returnCount (TNamed _ t) = returnCount t 
returnCount _ = 1
```

consider

```
object Test {                                                                                                  
    code {                                          
      function bad(x : Void{unit}) -> word {
        let r : word
        assembly {
          let junk := x    -- returnCount (TNamed "Void" TUnit) = 1, so
                           -- length [junk] = 1 == 1, type checker passes.                                       
                           -- Should fail: unit has 0 return slots.                                              
          r := 0                                                                                                 
        }                                                                                                        
        return r                                                                                                 
      }                                             
    }    
  }
```                
and, concversely
```
  object Test {                                                                                                  
    code {                                          
      function ok(p : Pair{word * word}) -> word {
        let r : word                                                                                             
        assembly {
          let a, b := p    -- returnCount (TNamed "Pair" (TPair TWord TWord)) = 1,                               
                           -- but length [a, b] = 2, so checker rejects this.                                    
                           -- Should pass: the pair occupies 2 slots.                                            
          r := add(a, b)                                                                                         
        }                                                                                                        
        return r                                    
      }  
    }
  } 
```

src/Language/Hull/TypeCheck.hs
```diff
@@ -0,0 +1,298 @@
...
+lookupAsmFun f = do
+  funs <- gets hull_funs
+  let s = show f
+      key = fromMaybe s (stripPrefix "usr$" s)
+  case Map.lookup key funs of
+    Just sig ->
+      pure $
+        HullFunSig
+          (map (const TWord) (hsig_args sig))
+          (if typeEq (hsig_ret sig) TUnit then TUnit else TWord)
```
mbenke: Might be better to use nReturns, expecially if we want to call Hull/Sail from Yul in the future:


```suggestion
          (nReturns (returnCount (hsig_ret sig)))  
```

mbenke (approved):
Well done, just some minor issues.

mbenke:
Actually there is a problem with this weird code:

```
object Unsafe {
  code {
    function fromWord$unit (x : word) -> unit {
      let result : unit
      assembly {
        result := x
      }
      return result
    }
    function main () -> word {
      fromWord$unit(7)
      return 42
    }
  }
}
```

in `checkAsmStmt`:
- `mapM_ (lookupVar . show) ns` verifies result exists and retrieves its type TUnit, but immediately discards it. Only existence is checked. 
- `nExpected = length ns = 1` This counts variable names, not slots. result : unit occupies 0 slots, so the expected count should be returnCount TUnit = 0. But the code never looks at the LHS types.                                                                                                              - `nActual = returnCount TWord = 1.`

src/Language/Hull/TypeCheck.hs
```diff
@@ -0,0 +1,298 @@
...
+  t <- checkAsmExp e
+  let nExpected = length ns
+      nActual = returnCount t
+  when (nActual /= nExpected) $
+    hullError $
+      unlines
+        [ "Return count mismatch in assignment",
+          "  assigning  " ++ show nExpected ++ " variable(s)",
+          "  expression returns " ++ show nActual ++ " value(s)"
+        ]
```
mbenke, line 215: count stack slots rather than variables:

```suggestion
checkAsmStmt (YAssign ns e) = do
  lhsTypes <- mapM (lookupVar . show) ns
  t <- checkAsmExp e
  let nExpected = sum (map returnCount lhsTypes)
      nActual = returnCount t
  when (nActual /= nExpected) $
    hullError $
      unlines
        [ "Return count mismatch in assignment",
          "  assigning  " ++ show nExpected ++ " slot(s)",
          "  expression returns " ++ show nActual ++ " value(s)"
        ]
```

PR #369, opened 2026-05-05
Increase the solver iteration limit

rodrigogribeiro:
Simple increase in solver iteration limit allows the compilation of the deposit contract without issues.


Y-Nak (approved):
LGTM. Probably we should need to revisit this, but it should be fine for now.

PR #371, opened 2026-05-06
Specialise: better diagnostics for unresolved type variables

mbenke:
This PR introduces better diagnostics in Specialise so that it should not PANIC so much anymore. 
Sadly does not fix problems with unresolved type variables, but should make for better UX.
`test/examples/cases/monomorphic-require.solc` suggests a way of implementing require.


rodrigogribeiro (approved):
LGTM.

PR #372, opened 2026-05-07
Specialise: in StmtExp, ground all type vars to ()

mbenke:
If we have an statement consisting of an expression, such as `revert(0,0);` the value of the expression is dropped anyway, so it should be safe to ground all variables in its type to ().

While this does not solve the problem of unresolved variables at specialisation time, at least it should solve problems with `revert` and hence `require`


rodrigogribeiro (approved):
LGTM.

PR #373, opened 2026-05-08
Remove unused svgs

Y-Nak:
This PR just removes unused SVGs, which causes a problem on macOS because of its case-insensitive file system.


rodrigogribeiro (approved):
Thanks for noticing that!

PR #374, opened 2026-05-08
Skolem let fix

rodrigogribeiro:
This PR fixes an issue in type checking Yul code and adds a test case suggested by @mbenke.

### DIFF
--- a/test/Cases.hs
+++ b/test/Cases.hs
@@ -429,8 +429,6 @@ cases =
       runTestForFile
         "multi-stmt-var-leaf.solc"
         caseFolder,
-      runTestForFile "simpleDiscount.solc" caseFolder,
-      runTestExpectingFailure "skolem-let.solc" caseFolder
     ]
   where
     caseFolder = "./test/examples/cases"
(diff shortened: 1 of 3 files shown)


mbenke:
I have some doubts about the error message:

```
$ esolc test/examples/cases/skolem-let.solc 
test/examples/cases/skolem-let.solc
Type not polymorphic enough! The annotated type is:
forall a . word -> a
but the infered type is:
word -> word
in:
forall a . function fromWord (x : word) -> a

 - in:forall a . function fromWord (x : word) -> a {
   let result ;
   assembly {
      result := x
   }
   return result;
}
```

...but x is in fact declared to have type a (which is misrepresented in the error message) and the problem is in `result := x` but I see it may be hard to report in a better way ("types $123 and word do not unify" probably would be even less helpful)

test/Cases.hs
```diff
@@ -428,7 +429,8 @@ cases =
       runTestForFile
         "multi-stmt-var-leaf.solc"
         caseFolder,
-      runTestForFile "simpleDiscount.solc" caseFolder
+      runTestForFile "simpleDiscount.solc" caseFolder,
+      runTestExpectingFailure "skolem-let.solc" caseFolder
```
mbenke: This two tests are duplicates (appear earlier on lines 333 and 423)

```suggestion
```

mbenke (approved):
Well done. Just minor complaints (feel free to ignore my rant about error message).

PR #378, opened 2026-05-11
Add helpers for `revert` and custom errors with `Error` type

axic:


rodrigogribeiro (approved):
Good work!

std/std.solc
```diff
@@ -142,10 +146,13 @@ function log1(v:t, topic:word) -> () {
 }
 
 function unimplemented() -> () {
-  assembly {
-    mstore(0, 0xaddc0de1)
-    revert(0, 4)
```
axic, line 147: Also note all of these did return `0x00000000` in the past, because words are big endian in memory, and thus `revert(28, 4)` would have positioned it correctly.

PR #382, opened 2026-05-14
Group basic types and implementation in std for clarity

axic:


std/std.solc
```diff
@@ -743,6 +743,12 @@ data bytes;
 // TODO: should this be a typedef over `bytes`?
 data string;
 
+instance string:Add {
```
rodrigogribeiro, line 746: I do not like the idea of overloading the `+` operator for strings. Maybe, we should use other operator for string concatenation.

axic: Please note I didn't add anything in this PR, just moved things around. This is actually used by the dispatch.

rodrigogribeiro: Ohhh... I didn't notice that there was this instance on the std library. I'll discuss this on the meeting next week.

rodrigogribeiro (approved):
LGTM.

Issue #380, opened 2026-05-14
Weird bug with assembly and variable access


axic:
Found this in #378 (https://github.com/argotorg/solcore/actions/runs/25882316805/job/76065173026?pr=378).

`do_exec` in dispatch has this:
```
    // check we have enough calldata for the head of args
    let hdsz = ABIAttribs.headSize(pargs);
    assembly {
        if lt(sub(calldatasize(), 4), hdsz) {
            revert(0,0);
        };
    }
```

This change still works:
```
    // check we have enough calldata for the head of args
    let hdsz = ABIAttribs.headSize(pargs);
    let cdsz: word;
    assembly {
      cdsz := calldatasize()
    }
    assembly {
        if lt(sub(cdsz, 4), hdsz) {
            revert(0,0);
        };
    }
```

However this one fails:
```
    // check we have enough calldata for the head of args
    let hdsz = ABIAttribs.headSize(pargs);
    let cdsz: word;
    assembly {
      cdsz := calldatasize()
    }
    if ((cdsz - 4) < hdsz) {
      assembly {
        revert(0, 0)
      }
    }
```

The error output is the following (it fails on two test cases, monomorphic-require and polymorphic-require):
```
    polymorphic-require.solc:                                                              FAIL (1.13s)
      test/Cases.hs:457:
      Undefined name: lt
       - in:lt
       - in:invokable.invoke(lt, (Sub.sub(cdsz, 4), hdsz))
       - in:forall payability args rets fn .
      fn : invokable (args, rets), args : ABIAttribs, rets : ABIAttribs, ABIDecoder(args, CalldataWordReader) : ABIDecode (args), rets : ABIEncode, payability : MethodLevelCallvalueCheck => function std.dispatch.$impl$do_exec (ppayability : Proxy(payability), pargs : Proxy(args), prets : Proxy(rets), fn : fn) -> () {
         MethodLevelCallvalueCheck.checkCallvalue(Proxy.Proxy : Proxy(payability));
         let hdsz = ABIAttribs.headSize(pargs) ;
         let cdsz : word ;
         assembly {
            cdsz := calldatasize()
         }
         if (invokable.invoke(lt, (Sub.sub(cdsz, 4), hdsz))) {
            assembly {
               revert(0, 0)
            }
         } else {
         }
         let ptr : calldata(bytes) = calldata.calldata(4) ;
[...]

PR #389, opened 2026-05-15
Turn Error into an ADT and add require/assert

axic:


std/std.solc
```diff
@@ -80,6 +81,7 @@ export {
   revert,
   revertEmpty,
```
axic, line 82: Considering removing this, since we can use `revert(Error.Empty)`.

test/imports/import_std_minimal.solc
```diff
@@ -1,3 +1,4 @@
 import std;
+import std.{not};
```
axic: All these workarounds due to #380, if the fix for that is merged first then we can omit these changes here.

axic: Dropped and rebased.

test/examples/dispatch/miniERC20.solc
```diff
@@ -73,12 +73,12 @@ contract MiniERC20 {
 
   function transferFrom(src:address, dst:address, amt:uint256) -> bool {
      let msg_sender = caller();
-     require( balances[src] >= amt /* "token/insufficient-balance" */
+     myrequire( balances[src] >= amt /* "token/insufficient-balance" */
```
mbenke: Perhaps it would make sense to exercise the new Error system here - if I understand it correctly then this should be 
```
require( balances[src] >= amt /* "token/insufficient-balance" */
        , Error.Error(0x746f6b656e2f696e73756666696369656e742d62616c616e6365)
        )
```
Probably no need to change it in other tests, but miniERC20 is one of our main ones.

axic: The error encoding follows the ABI, so it is `sha256(<errorname>(<arguments>))[:4]` (truncating to the first 4 bytes). Since error and function names cannot have forward slashes (`/`), the original error `token/insufficient-balance` cannot be represented. This is the reason I haven't changed it.

We need to change it to something like `InsufficientBalance()` if we follow Solidity's model.

axic: @mbenke updated the dispatch tests

test/examples/dispatch/miniERC20.solc
```diff
@@ -73,14 +65,10 @@ contract MiniERC20 {
...
-     require( balances[src] >= amt /* "token/insufficient-balance" */
-            , 0x746f6b656e2f696e73756666696369656e742d62616c616e6365
-	    );
+     require(balances[src] >= amt, Error(0xf4d678b8)); // InsufficientBalance()
 
      if (src != msg_sender && allowance[src][msg_sender] != (Num.maxVal():uint256)) {
-        require( allowance[src][msg_sender] >= amt /* "token/insufficient-allowance" */
-	       , 0x746f6b656e2f696e73756666696369656e742d616c6c6f77616e6365
-	       );
+        require(allowance[src][msg_sender] >= amt, Error(0xf4d678b8)); // InsufficientBalance()
```
mbenke: Copy-paste error - this should be InsufficientAllowance


```suggestion
        require(allowance[src][msg_sender] >= amt, Error(0x13be252b)); // InsufficientAllowance()

```

mbenke (approved):
Minor copy/paste bug otherwise well done. 

I plan to add a `runsol.sh` modification to display symbol errors:

```
./runsol.sh test/examples/dispatch/miniERC20.solc \
    --create-arguments "f(string,string,uint256)" "Token" "TKN" "1000" \
    --runtime-sender "0x000000000000000000000000000000000000cafe" \
    --runtime-calldata "transferFrom(address,address,uint256)" \
      "0x000000000000000000000000000073656e646572" \
      "0x000000000000000000000000000000000000dead" \
      "100"
Processing: test/examples/dispatch/miniERC20.solc
Compiling to hull...
Emitting hull for contract MiniERC20
Writing to output1.hull
Generating Yul...
writing output to /home/ben/work/review/build/miniERC20.yul
Compiling to bytecode...
Hex output: /home/ben/work/review/build/miniERC20.hex
Executing...
Creation successful
returndata: 0x61000b36...6107b556
Executing...
Execution failed: execution reverted
returndata: 0xf4d678b8
error selector: InsufficientBalance()
```

PR #391, opened 2026-05-15
Add basic version of concatenation

axic:
This adds the following features:
1. `concat(a, b) -> memory(bytes)`, which takes two supported inputs and encodes it into a new memory array
2. `to_bytes(a) -> memory(bytes)`, which does the same but for a single input
3. the type `empty(size)`, which represent a `size`-long memory array filled with the value 0 
4. the `MemoryAttributes` class which enables these features (currently implemented for `bytes32`, `memory(bytes)`, `empty(n)`)

Example:
```
concat(bytes32(0x1234), empty(12))  # output is 44 bytes
```


std/std.solc
```diff
@@ -1878,3 +1878,56 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    // The size needed for the value.
+    function len(v: t) -> word;
+    // If this is a non-memory type, memory is allocated first.
+    function ptr(v: t) -> word;
+    // Serialize the entire contents at a provided memory area.
+    function encodeInto(v: t, target: word) -> ();
+}
+
+// TODO: support variadic arguments
+forall a b . a:MemoryAttributes, b:MemoryAttributes => function concat(x: a, y: b) -> memory(bytes) {
```
axic, line 1969: This is here for discussion. It works in the deposit contract properly.

std/std.solc
```diff
@@ -1878,3 +1878,56 @@ forall i a . i:Typedef(word), a:StorageType  =>
 function ridx( m: storage(mapping(i,a)), x:i) -> a {
   return StorageType.sload(hash2(Typedef.rep(m), Typedef.rep(x)));
 }
+
+// Concatenation
+
+
+forall t . class t:MemoryAttributes {
```
axic, line 1959: There's some overlap with ABI encoding. I think in the long term there could be a generalized trait and then three specializations: ABI (calldata/returndata), storage, (runtime) memory.

std/std.solc
```diff
@@ -1878,3 +1878,56 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    // The size needed for the value.
+    function len(v: t) -> word;
+    // If this is a non-memory type, memory is allocated first.
+    function ptr(v: t) -> word;
```
axic, line 1963: This is not actually needed yet, could remove for now.

std/std.solc
```diff
@@ -1878,3 +1878,105 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    }
+    function encodeInto(v: memory(bytes), target: word) -> () {
+        let v_ = Typedef.rep(v);
+        assembly {
+            mcopy(target, add(v_, 32), mload(v_))
+        }
+    }
+}
+
+forall a . a:MemoryAttributes => function to_bytes(x: a) -> memory(bytes) {
```
axic, line 2009: This allows creating `memory(bytes)` of any supported type. If there was variadic input to `concat`, one could consider this as the single-argument specialization.

std/std.solc
```diff
@@ -1878,3 +1878,105 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    // The size needed for the value.
+    function len(v: t) -> word;
+    // If this is a non-memory type, memory is allocated first.
+    function ptr(v: t) -> word;
+    // Serialize the entire contents at a provided memory area.
+    function encodeInto(v: t, target: word) -> ();
+}
+
+// TODO: support variadic arguments
+forall a b . a:MemoryAttributes, b:MemoryAttributes => function concat(x: a, y: b) -> memory(bytes) {
```
axic, line 1969: I know that this could be done with operators (such as Add), but that will always have two operands. Here I hope we could have variadic inputs and thus reducing the memory waste when dealign with 2+ inputs.

std/std.solc
```diff
@@ -1878,3 +1878,105 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    let res = allocate_memory(32 + len);
+    mstore(res, len);
+    MemoryAttributes.encodeInto(x, res + 32);
+    return memory(res);
+}
+
+// Placeholder for an empty memory area.
+// The value is the size of the area in bytes. The area will be zeroed upon serialization.
+// NOTE: not implementing Typedef by design.
+data empty = empty(word);
```
axic, line 2020: This is a lightweight abstraction to avoid allocating memory. It only gets allocated as part of concat or to_bytes.

std/std.solc
```diff
@@ -1878,3 +1878,105 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+        // Zero out 32-byte words
+        let words := div(len, 32)
+        for { let i := 0 } lt(i, words) { i := add(i, 1) } {
+            mstore(ptr, 0)
+            ptr := add(ptr, 32)
+        }
+
+        // Zero out trailing bytes
+        for { } lt(ptr, end_ptr) { ptr := add(ptr, 1) } {
+            mstore8(ptr, 0)
```
axic: If there's the dedicated zero slot (0x60-0x7f in classic solidity) then this could be an mcopy instead:
```
mcopy(ptr, 0x60, sub(end_ptr, ptr))
```

axic: There is, will update.

std/std.solc
```diff
@@ -1878,3 +1878,105 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    }
+}
+
+forall a . a:MemoryAttributes => function to_bytes(x: a) -> memory(bytes) {
+    let len = MemoryAttributes.len(x);
+    let res = allocate_memory(32 + len);
+    mstore(res, len);
+    MemoryAttributes.encodeInto(x, res + 32);
+    return memory(res);
+}
```
mbenke, line 2015: Why not just `return memory(MemoryAttributes.ptr(x));`? 
For `memory(bytes)` this creates a copy, is this intentional?

axic:
@mbenke should we merge this and then you can build your version on top of this reusing some of the helpers (empty/zeroize/...)?

std/std.solc

std/std.solc
mbenke: Here is my idea of how you can implement a "variadic" function (actually accepting a tuple of any size). Currently it runs out of gas, but so does nested concat (nested concat is probably the wrong way to go, but we can improve performance later).
This copies most of your code since it is not exported in std.

```
import std.{Typedef,memory,bytes,bytes32,allocate_memory,mstore,mload};

pragma no-patterson-condition Catenable;

forall t.class t:Catenable {
  function cat(x:t) -> memory(bytes) ;
}

forall t. t:MemoryAttributes =>
default instance t:Catenable {
  function cat(x:t) -> memory(bytes) { to_bytes(x) }
}

forall a b . a:MemoryAttributes, b:Catenable =>
instance (a,b):Catenable {
  function cat(p:(a,b)) -> memory(bytes) {
    match p { | (x,y) =>  return concat(x:a,Catenable.cat(y)); }
  }
}

contract Cat {
   function main() -> word {
     let e : empty = empty(14);
     
     // let b : memory(bytes) = concat(e,concat(e,to_bytes(e))); // this runs out of gas too
     let b : memory(bytes) = Catenable.cat((e,e,e)); 
     return MemoryAttributes.len(b);
   }
}

// ------------------------------------------------------------------
// copied from #391

forall t . class t:MemoryAttributes {
    // The size needed for the value.
    function len(v: t) -> word;
    // If this is a non-memory type, memory is allocated first.
    function ptr(v: t) -> word;
    // Serialize the entire contents at a provided memory area.
    function encodeInto(v: t, target: word) -> ();
}


forall a b . a:MemoryAttributes, b:MemoryAttributes => function concat(x: a, y: b) -> memory(bytes) {
    let x_len = MemoryAttributes.len(x);
    let y_len = MemoryAttributes.len(y);
    let res: word = allocate_memory(32 + x_len + y_len);
    mstore(res, x_len + y_len);
    MemoryAttributes.encodeInto(x, res + 32);
    MemoryAttributes.encodeInto(y, res + 32 + x_len);
    return memory(res);
}

instance bytes32:MemoryAttributes {
    function len(v: bytes32) -> word {
        return 32;
    }
    function ptr(v: bytes32) -> word {
        let ptr = allocate_memory(64);
        mstore(ptr, 32);
        mstore(ptr + 32, Typedef.rep(v));
        return Typedef.rep(ptr);
    }
    function encodeInto(v: bytes32, target: word) -> () {
        mstore(target, Typedef.rep(v));
    }
}

instance memory(bytes):MemoryAttributes {
    function len(v: memory(bytes)) -> word {
        return mload(Typedef.rep(v));
    }
    function ptr(v: memory(bytes)) -> word {
        return Typedef.rep(v);
    }
    function encodeInto(v: memory(bytes), target: word) -> () {
        let v_ = Typedef.rep(v);
        assembly {
            mcopy(target, add(v_, 32), mload(v_))
        }
    }
}

forall a . a:MemoryAttributes => function to_bytes(x: a) -> memory(bytes) {
    let len = MemoryAttributes.len(x);
    let res = allocate_memory(32 + len);
    mstore(res, len);
    MemoryAttributes.encodeInto(x, res + 32);
    return memory(res);
}


// Placeholder for an empty memory area.
// The value is the size of the area in bytes. The area will be zeroed upon serialization.
// NOTE: not implementing Typedef by design.
data empty = empty(word);

instance empty:MemoryAttributes {
    function len(v: empty) -> word {
        match v {
            | empty(size) => return size;
        }
    }
    function ptr(v: empty) -> word {
        return 0; // TODO
    }
    function encodeInto(v: empty, target: word) -> () {
        let size;
        match v {
            | empty(size_) => size = size_;
        }
        zeroize_memory(target, size);
    }
}

function zeroize_memory(ptr: word, len: word) -> () {
    assembly {
        let end_ptr := add(ptr, len)

        // Zero out 32-byte words
        let words := div(len, 32)
        for { let i := 0 } lt(i, words) { i := add(i, 1) } {
            mstore(ptr, 0)
            ptr := add(ptr, 32)
        }

        // Zero out trailing bytes
        for { } lt(ptr, end_ptr) { ptr := add(ptr, 1) } {
            mstore8(ptr, 0)
        }
    }
}
```

std/std.solc
```diff
@@ -1878,3 +1878,103 @@ forall i a . i:Typedef(word), a:StorageType  =>
...
+    // The size needed for the value.
+    function len(v: t) -> word;
+    // If this is a non-memory type, memory is allocated first.
+    function ptr(v: t) -> word;
+    // Serialize the entire contents at a provided memory area.
+    function encodeInto(v: t, target: word) -> ();
+}
+
+// TODO: support variadic arguments
+forall a b . a:MemoryAttributes, b:MemoryAttributes => function concat(x: a, y: b) -> memory(bytes) {
```
mbenke, line 1969: Repeated concatenation such as
```
     let e : empty = empty(14);     
     let b : memory(bytes) = concat(e,concat(e,to_bytes(e))); 
```
seems to run out of gas.

std/std.solc
axic: Yes, `to_bytes` creates a copy by design.

std/std.solc
axic: Oh I missed this comment.

> (nested concat is probably the wrong way to go, but we can improve performance later).

My entire need for variadic input was to avoid nested concat, because concat is allocating memory and copying, and want to avoid duplicating memory on every iteration.

axic:
Pushed tests.

std/std.solc
axic: @mbenke test seem to succeed with this. Also the deposit contract uses nested concats and it works too. Are you sure it runs out of gas (at what limit)?

mbenke: I am getting this:
```
$ ./runsol.sh tmp/cat.solc
Processing: tmp/cat.solc
Compiling to hull...
Emitting hull for contract Cat
Writing to output1.hull
Generating Yul...
writing output to /home/ben/work/review/build/cat.yul
Compiling to bytecode...
Hex output: /home/ben/work/review/build/cat.hex
Executing...
Creation successful
returndata: 0x60405160...5b605d56
Executing...
Execution failed: gas uint64 overflow
returndata: 0x
```
for

```
contract Cat {
   function main() -> word {
     let e : empty = empty(14);
     
     let b : memory(bytes) = concat(e,concat(e,to_bytes(e))); 
     return MemoryAttributes.len(b);
   }
}
```

axic: @mbenke This seems to be some miscompilation bug. If I add this to contact.json then every single test fails. Renaming it to `main2` fixes it. Also note that `ABIEncode` is not implemented for `word` (after renaming it to main2 I was reminded), so my guess is `contract C { function main() -> word {} }` conflicts with the dispatcher and miscompiles something.

axic: Added #440 as a discussion, apart from the miscompilation bug.

mbenke: I do not see #440 as relevant here, but I think I see what you mean - the looping seems to be only loosely related.

mbenke (approved):
Tracking the looping problem as #442, but otherwsie this should be good to go.

std/std.solc
axic: Having the `main` function replaces dispatch and causes some miscompilation, thats why you have that out of gas, not because of concat.

PR #392, opened 2026-05-16
Tabled typeclass resolution

Y-Nak:
This PR implements [Tabled Typeclass Resolution](https://arxiv.org/abs/2001.04301).
The new resolver can be invoked via `--type-class-resolution tabled`. The default resolver is the existing one for compatibility.


src/Solcore/Frontend/TypeInference/TcResolution.hs
```diff
@@ -0,0 +1,918 @@
...
+  lift $ putSubst consumerSubst
+  lift $ info [">> Consuming (tabled):", pretty goals0]
+  goals1 <- lift $ elimEqualities goals0
+  goals <- lift $ withCurrentSubst goals1
+  case goals of
+    [] -> addAnswer target residualAcc
+    (p : ps) ->
+      case p of
+        InCls {} -> newSubgoal target residualAcc p ps
+        t1 :~: t2 -> do
```
mbenke: Is this case needed? We seem to eliminate equalities couple of lines above

Y-Nak: Oh, right. Thanks!

Y-Nak:
I also fixed one more bug in https://github.com/argotorg/solcore/pull/392/changes/96e7ff27f478bf74aebfe1d1ae20f966e396ae94

mbenke (approved):
Nice work!

PR #393, opened 2026-05-17
Making type annotations on class function signatures obrigatory.

### LINKED ISSUE #390: Mandatory return type check is missing on classes
The class definition on its own is valid (no compiler error), until an instance tries to implement it:
```
forall t . class t:Memory {
    function encodeInto(v: t, target: word);
}

instance bytes32:Memory {
    function encodeInto(v: bytes32, target: word) {}
}
```

When trying to implement it, the produced error is this:
```
Compiling to Hull...
Instance methods must have complete type signatures:
function encodeInto (v : bytes32, target : word)

 - in:instance bytes32 : Memory {
   function encodeInto (v : bytes32, target : word) {
   }
}
Error: sol-core compilation failed
```

Of course  adding `-> ()` to both fixes it.

rodrigogribeiro:
Fixes #390.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -1225,7 +1225,7 @@ fullSignature :: Signature Name -> TcM ()
 fullSignature sig =
   unless
     (isFullyAnnotated sig)
-    (throwError $ unlines ["Instance methods must have complete type signatures:", pretty sig])
+    (throwError $ unlines ["Top-level functions must have complete type signatures:", pretty sig])
 
 requireAnnotations :: FunDef Name -> TcM ()
 requireAnnotations (FunDef sig _) =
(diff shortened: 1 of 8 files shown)


src/Solcore/Frontend/TypeInference/TcStmt.hs
```diff
@@ -1225,7 +1225,7 @@ fullSignature :: Signature Name -> TcM ()
 fullSignature sig =
   unless
     (isFullyAnnotated sig)
-    (throwError $ unlines ["Instance methods must have complete type signatures:", pretty sig])
+    (throwError $ unlines ["Top-level functions must have complete type signatures:", pretty sig])
```
mbenke: Why this change? It seems that this message is used for class methods and subsequent `requireAnnotations` for top level functions. 
It might be also good to give the error functions better names (I know, I know).

Right now my errroneous code reports

```
Top-level functions must have complete type signatures:
function cat (t) -> word

 - in:forall t . class  t : Catenable {
   function cat (t) -> word;
}
```

```suggestion
    (throwError $ unlines ["Class and instance methods must have complete type signatures:", pretty sig])
```

mbenke (approved):
One minor suggestion for error messages. Otherwise, well done.

PR #394, opened 2026-05-17
Removing LHS from a mstore assignment.

### LINKED ISSUE #383: Return stack count is ignored in assembly
This example in `std.solc` compiles:
```
        | 0 => assembly { res := mstore(add(loc, mul(iw, 32)), vw); }
```

`mstore` doesn't return a stack item. I imagine the output Yul has it correctly removing the assignment.

rodrigogribeiro:
Fixes #383, by removing an unnecessary assignment in std.


axic:
This code is now gone, so the PR can be closed.

PR #400, opened 2026-05-20
Comptime integers - first PoC

mbenke:
First PoC of comptime integers 

Integer literals are still of type `word`, eventually they should be of type `integer` but this is a big change (in the sense it may affect almost all tests)

Tests/examples (more to come):
- `test/examples/comptime/integer-lit.solc`
- `test/examples/comptime/integer-lit.solc`


rodrigogribeiro (approved):
Great work!

PR #404, opened 2026-05-22
Add bitwise assembly helpers

axic:
These can be used to rewrite more of the assembly blocks.


axic:
Coming to think of it, #405 is a better approach in my opinion.

rodrigogribeiro (approved):
LGTM.

PR #405, opened 2026-05-22
Add `std/opcodes`

axic:
The goal here is to reduce ad-hoc helpers. The `std/opcodes.solc` file could be auto-generated, following fixed templates.

### DIFF
--- a/std/opcodes.solc
+++ b/std/opcodes.solc
@@ -1,14 +1,84 @@
 export {
     stop,
     add,
-    mstore,
-    mload,
-    mcopy,
-    sstore,
-    sload,
+    mul,
+    sub,
+    div,
+    sdiv,
+    mod,
+    smod,
+    addmod,
+    mulmod,
+    exp,
+    signextend,
+    lt,
+    gt,
+    slt,
+    sgt,
+    eq,
+    iszero,
+    and,
+    or,
+    xor,
+    not,
+    byte,
+    shl,
+    shr,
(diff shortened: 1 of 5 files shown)


std/opcodes.solc
```diff
@@ -0,0 +1,43 @@
...
+
+function callvalue() -> word {
+    let res;
+    assembly {
+        res := callvalue()
+    }
+    return res;
+}
+
+function revert(a: word, b: word) -> () {
```
axic, line 668: This will likely conflict with special handling for `revert(string literal)`.

std/std.solc
```diff
@@ -1088,26 +1077,34 @@ instance bool:ABIEncode {
...
     }
+    return res;
+}
+
+function round_up_to_mul_of_32(value:word) -> word {
+    return and_(value + 31, not_(31));
+}
 
-    return tail;
+function encodeIntoFromBytesLike(srcPtr:word, basePtr:word, offset:word, tail:word) -> word {
```
axic, line 1197: As an example using the opcode helpers assembly blocks can be rewritten line-by-line, and after that it is easy to simplify it too.

axic:
Will split this into two parts:
1) This PR adding the script and `std/opcodes.solc`.
2) A follow up PR using the helpers from it.

axic:
@rodrigogribeiro this should be ready. Moved the usage to #449 (blocked by #435)

PR #410, opened 2026-05-23
Add `address:ABIEncode`

axic:

PR #413, opened 2026-05-23
Return error code upon hitting fallback

axic:


std/dispatch.solc
```diff
@@ -289,6 +289,8 @@ forall methods fallback . methods:RunDispatch, fallback:ExecMethod => instance C
   }
 }
 
-function revert_handler() -> () {
-  assembly { revert(0,0) }
+// This is the default fallback used if none is defined.
+function fallback_default_implementation() -> () {
+  let NoSelectorMatchedWithoutFallback = Error(0x4924aef0);
+  revertWithError(NoSelectorMatchedWithoutFallback);
```
axic, line 295: Took me a while debugging #410 because I forgot to change the selector in the tests. As previously, returning a proper error fixes this problem.

test/examples/dispatch/basic.json
```diff
@@ -69,6 +69,18 @@
           "returndata":"00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000004deadbeef00000000000000000000000000000000000000000000000000000000",
           "status": "success"
         }
+      },
+      {
+        "input": {
+          "text-calldata": "missing fallback",
```
axic, line 111: I tried creating a test for defined fallback too, but there's no way to define it.

src/Solcore/Desugarer/ContractDispatch.hs
```diff
@@ -75,7 +75,7 @@ genMainFn addMain (Contract cname tys cdecls)
         [ proxyExp (TyCon "NonPayable" []),
           proxyExp unit,
           proxyExp unit,
-          Var "revert_handler"
+          Var "fallback_default_implementation"
```
axic, line 78: Also wonder if this could be implemented directly in disaptch.solc

mbenke (commented):
Nice! I have added `NoSelectorMatchedWithoutFallback()` to the 4bytes db so that `./runsol.sh` now prints it when encountered:
```
$ ./runsol.sh test/examples/dispatch/basic.solc  --runtime-calldata "noSuchMethod()"
...
Execution failed: execution reverted
returndata: 0x4924aef0
error selector: NoSelectorMatchedWithoutFallback()
```

mbenke (approved):

PR #415, opened 2026-05-26
Fix too strict import rule

### LINKED ISSUE #412: Import statements have strict ordering rules (and/or the parser is very strict)
Import statements **must** be placed before exports **and** pragmas.

Otherwise the parser spits out an error:
```
Parse error while processing lexeme: TImport
 at line 4, column 1
```

Y-Nak:
Fixes #412.


mbenke (approved):
Well done.

PR #416, opened 2026-05-26
Improve diagnostics

Y-Nak:
Improve compiler diagnostics.
This PR adds structured diagnostics and source spans in more places, so errors can show file, line, column and labels.

Remaining things that should be addressed later are:
* some old `tcmError` / `SC0299` errors are still there
* some fallback span guessing is still used
* location tests are not perfect, they do not check every span is preserved exactly


mbenke (commented):
In some cases the error message is repeated, e.g.

```
$ esolc tmp/unsafeFromWord.solc 
tmp/unsafeFromWord.solc
error: module typecheck failed for /home/ben/work/review/tmp/unsafeFromWord.solc (no desugaring):
Type not polymorphic enough! The annotated type is:
forall a . word -> a
but the infered type is:
word -> word
in:
forall a . function fromWord (x : word) -> a

 - in:forall a . function fromWord (x : word) -> a {
   let result ;
   assembly {
      result := x
   }
   return result;
}
  ──> /home/ben/work/review/tmp/unsafeFromWord.solc:3:30
  │
3 │       assembly { result := x } 
  │                              ^ module typecheck failed for /home/ben/work/review/tmp/unsafeFromWord.solc (no desugaring):
Type not polymorphic enough! The annotated type is:
forall a . word -> a
but the infered type is:
word -> word
in:
forall a . function fromWord (x : word) -> a

 - in:forall a . function fromWord (x : word) -> a {
   let result ;
   assembly {
      result := x
   }
   return result;
}
```

src/Solcore/Frontend/Module/Loader.hs
```diff
@@ -153,29 +157,60 @@ visit cfg moduleId sourcePath = do
...
+moduleReferenceDiagnostic code sourcePath refKind modulePath message =
+  loaderDiagnosticWithLabels
+    code
+    message
+    (maybe [] pure (primaryModulePathLabel (moduleReferenceLabelMessage code refKind) modulePath))
+    [sourcePath, refKind ++ " " ++ Mod.modulePathDisplay modulePath]
+    (moduleReferenceHelp code)
+
+moduleReferenceLabelMessage :: String -> String -> String
+moduleReferenceLabelMessage "SC0118" _ = "external library import"
```
rodrigogribeiro: Could this be an enumeration instead of a String?

rodrigogribeiro (approved):
The previous comment on the codes used is not a blocker: I just want to understand the logic behind the choice for the codes.

Great work!

mbenke (commented):
Good work.  Just resolve conflicts and let us merge this.

PR #418, opened 2026-05-27
Rename revert(string) to revertLit(string)

axic:


std/std.solc
```diff
@@ -162,7 +162,7 @@ function out_of_bounds() -> () {
 // ------------------------------------------------------------------
 // EmitHull has special handling for `revert("...")` after MastEval has
 // constant-folded the argument to a string literal.
-function revert(s:string) -> () {
+function revertLit(s:string) -> () {
```
axic, line 165: This follows the same naming as the other `*Lit` helpers. Goal is remove overuse of the name `revert`, and this can now be easily plugged into the ADT below *and* avoid conflict with the opcodes PR.

axic: Tbh not fully clear about the intention here, but I assume this exists as a wrapper so that the Haskell part of the compiler can replace it, if needed.

mbenke: No, the *Lit functions are meant to be evaluated at comptime. `revert` in the other hand does not. 

I also don't understand the need of messing with Hull.

axic: So what is this function then? It does say it has special casing in Hull.

mbenke: Yes, it passes the string literal to Yul, so you can do things like

```
import std.{revert,string};

contract Rev {
  function main() -> () {
    revert("foobar");
  }
}
```

axic: But it only takes string literals, and nothing else, correct? Feels like Lit is a good suffix because of that?

PR #419, opened 2026-05-27
Add `fallback` keyword for contract-level fallback functions

axic:
Introduces a new `fallback` keyword that defines an anonymous, optional function inside a contract — same shape as `function name(...)` but without a name and limited to zero or one per contract. The contract dispatch desugarer wires the user-defined fallback into the generated `Contract` value (replacing `fallback_default_implementation`) so it runs when no selector matches. The default implementation is retained when no fallback is declared.

Also renames the `fallback` type variables in dispatch.solc to `fb` since `fallback` is now a reserved keyword.

Testing cases:
- `basic.json` has the missing fallback case
- `fallback.json` has the non-payable fallback cases
- `payable.json` has the payable fallback cases

This was aided by Claude.

### DIFF
--- a/src/Solcore/Desugarer/ContractDispatch.hs
+++ b/src/Solcore/Desugarer/ContractDispatch.hs
@@ -15,7 +15,7 @@ module Solcore.Desugarer.ContractDispatch
 where
 
 import Data.List (mapAccumL)
-import Data.Maybe (mapMaybe)
+import Data.Maybe (fromMaybe, listToMaybe, mapMaybe)
 import Data.Set (Set)
 import Data.Set qualified as Set
 import Language.Yul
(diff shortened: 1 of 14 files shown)


std/dispatch.solc
```diff
@@ -27,7 +27,7 @@ pragma no-bounded-variable-condition ;
 
 // A contract contains a tuple of methods and a single fallback
 // TODO: implement receive()
-data Contract(methods, fallback) = Contract(methods,fallback);
+data Contract(methods, fb) = Contract(methods,fb);
```
axic, line 30: ~~The renaming is optional.~~

It would conflict with the keyword.

src/Solcore/Desugarer/ContractDispatch.hs
```diff
@@ -69,14 +83,23 @@ genMainFn addMain (Contract cname tys cdecls)
...
     methods = tupleExpFromList (fmap mkMethod (mapMaybe unwrapSigs cdecls))
-    fallback =
-      Con
-        "Fallback"
-        [ proxyExp (TyCon "NonPayable" []),
-          proxyExp unit,
-          proxyExp unit,
-          Var "fallback_default_implementation"
-        ]
+    fallback = case findFallback c of
```
axic, line 86: I think this logic now could be moved to disaptch.solc, have a match there.

src/Solcore/Frontend/Parser/SolcoreParser.y
```diff
@@ -358,6 +360,16 @@ Function : Signature Body {FunDef $1 $2}
 -- Proposed Rust-style short return, e.g `function d(x) { 2*x }`
          | Signature '{' Expr '}' {FunDef $1 [Return $3]}
 
+-- Fallback declaration: a nameless function. At most one per contract.
```
axic: This follows the same syntax as Classic Solidity:
```
fallback() -> () {}
```

vs

```
fallback() public external {}
```

One question we can debate whether it should allow different return types (and therefore use ABIEncoder). Classic Solidity afair doesn't allow that, one must encode manually. This PR so far is relaxed in that matter.

axic: And since there is already `constructor()` I'd argue this is the correct syntax.

axic:
@rodrigogribeiro also rebased this.

std/dispatch.solc
```diff
@@ -273,29 +273,27 @@ forall c . class c:RunContract {
...
+        // calldata shorter than 4 bytes can't contain a selector — skip
+        // dispatch and invoke the fallback directly (matches Solidity)
         let cdsz: word;
         assembly {
           cdsz := calldatasize()
         }
         match cdsz < 4 {
-          | true =>
-            let SelectorTooShort = Error(0x35ffc70e);
-            revertWithError(SelectorTooShort);
```
axic, line 291: This error code is now moot and is folded into `NoSelectorMatchedWithoutFallback`, which I think makes sense.

If we really want it could be complicated further to check if no fallback exists *and* input is too short, but the new logic here follows Solidity a bit closer.

test/examples/dispatch/payable.solc
```diff
@@ -19,4 +19,14 @@ contract PayableTest {
         }
         return uint256(value);
     }
+
+    payable fallback() -> () {
```
axic, line 23: This may be split into `receive` vs. `fallback` as in Solidity, but I think its better to have that as a separate discussion. `receive` is an extra syntactic/"security" sugar, which is optional, but without `fallback` a number of things can't be implemented.

axic:
The build failure is fixed by #428.

axic:
@rodrigogribeiro this is now ready.

rodrigogribeiro (approved):
LGTM.

PR #420, opened 2026-05-27
Remove legacy NumLib and tests for it

axic:
Follow up of #398


std/NumLib.solc
```diff
@@ -1,162 +0,0 @@
-import types;
```
axic, line 1: This was failing anyway, it is imported by IndexLib which only had a "runTestExpectingFailure" test case. It also shows those kind of "expect to fail" test are not very useful unless a specific error message is also expected.

test/examples/spec/114map.solc
axic, line 1: This file was not in Cases.hs

test/examples/spec/IndexLib.solc
axic, line 1: Also not in cases.

test/examples/cases/withdraw.solc
axic, line 1: The tests of this are also contained in 126nanoerc20.solc nad 127microerc20.solc.

mbenke (approved):

PR #422, opened 2026-05-30
comptime integers part 2

mbenke:
This is the second part of comptime integer implementation: literals are of type integer; desugaring adds `Int.fromInteger`


rodrigogribeiro (approved):
Great work!

PR #421, opened 2026-05-28
Fix module system bugs

Y-Nak:
Fixes #406, Fixes #411

PR #423, opened 2026-05-30
Cache module public interfaces

Y-Nak:
This PR fixes the performance regression that was introduced in #421.


rodrigogribeiro (commented):
Good work! As a suggestion, maybe we could take the opportunity to implement some kind of interface files: since typing is the most expensive activity in modules, we could store the type annotated tree in disk, and reuse its result if there's no change on the source file.

PR #425, opened 2026-06-02
Type-check Yul assignment arity against builtin return count

### LINKED ISSUE #383: Return stack count is ignored in assembly

axic:
Yul builtins/opcodes either return a single value or none at all (when typed as `unit`). The Yul type checker previously accepted `let x := mstore(1, 1)` and similar assignments even though `mstore` does not produce a value. Compare the number of names on the left-hand side of `YAssign`/`YLet` with the actual arity of the right-hand side and report a type error on mismatch.

Fixes #383.

Disclaimer: Mostly created by Claude.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -1547,7 +1547,7 @@ tcYulBlock (s : ss) =
     pure (ns ++ nss, t)
 
 tcYulStmt :: YulStmt -> TcM ([Name], Ty)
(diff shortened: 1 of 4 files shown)


rodrigogribeiro (approved):
Looks good! Great work!

axic:
> Great work!

Claude is pretty good! 😅

PR #428, opened 2026-06-02
Force UTF-8 locale in the compiler

axic:
This matches Solidity, which states the source code inputs are considered UTF-8.

### DIFF
--- a/test/examples/dispatch/miniERC20.solc
+++ b/test/examples/dispatch/miniERC20.solc
@@ -54,6 +54,7 @@ contract MiniERC20 {
     return totalSupply;
   }
 
+  // Note that this is not access guarded — the minting always goes to the owner
   function mint(amount:uint256) -> () {
     balances[owner] = Num.add(balances[owner], amount);
     totalSupply = Num.add(totalSupply, amount);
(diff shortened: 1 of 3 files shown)


test/examples/dispatch/miniERC20.solc
```diff
@@ -54,6 +54,7 @@ contract MiniERC20 {
     return totalSupply;
   }
 
+  // Note that this is not access guarded — the minting always goes to the owner
```
axic, line 57: The — is em-dash.

axic: As seen in https://github.com/argotorg/solcore/actions/runs/26844901955/job/79161668281?pr=428, the CI does not have locale set to UTF8 , we could rectify this with `export LC_ALL=C.UTF-8 (or en_US.UTF-8)`, but ultimately the compiler should be consistent.

```
       > Processing: test/examples/dispatch/miniERC20.json
       > Compiling to Hull...
       > sol-core: /build/nc8s6kw57bikym56fjcqhp0x3bz0jm9w-source/test/examples/dispatch/miniERC20.solc: hGetContents: invalid argument (cannot decode byte sequence starting from 226)
       > Error: sol-core compilation failed
```

rodrigogribeiro (approved):
LGTM.

PR #429, opened 2026-06-02
Resolve qualified constructor access on same-name-constructor types

### LINKED ISSUE #430: Cannot use other variants of `Error` in std
`require(true, Error.Empty)` fails with

```
Unqualified constructor:
Empty
Use Type.Constructor form.
```

axic:
When a data type T has a same-name constructor (e.g. `data Error = Error(word) | Empty | ...`), the inner resolver for the bare receiver `@Error@` in `@Error.Empty@` eagerly produced `@Con (QualName Error Error) []@`. The outer ExpVar/ExpName resolver only matched `@Just (Var d)@` for qualifier-style access, so it fell through to the catch-all and reported "Unqualified constructor: Empty. Use Type.Constructor form." — even though the user had used the qualified form.

Treat a same-name no-arg constructor receiver as if it were `@Var d@` before dispatching on the case table, restoring the qualifier-handling paths for `@Error.Empty@` and `@Error.Msg(x)@`.

Fixes #430

### DIFF
--- a/src/Solcore/Frontend/Syntax/NameResolution.hs
+++ b/src/Solcore/Frontend/Syntax/NameResolution.hs
@@ -510,6 +510,17 @@ resolveSameNameConstructorName n =
   where
     leaf = constructorLeafName n
 
+-- A receiver like @Error@ in @Error.Empty@ is first parsed as an expression
+-- on its own and resolved before the outer member-access context is known.
+-- When the type has a same-name constructor (@data Error = Error(...) | ...@),
+-- the inner resolver eagerly produces @Con (QualName Error Error) []@ for the
+-- bare name. Treat that shape as if it were @Var Error@ when it appears in
+-- qualifier position so the outer qualifier-handling cases still match.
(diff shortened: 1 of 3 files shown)


src/Solcore/Frontend/Syntax/NameResolution.hs
```diff
@@ -510,6 +510,17 @@ resolveSameNameConstructorName n =
   where
     leaf = constructorLeafName n
 
+-- A receiver like @Error@ in @Error.Empty@ is first parsed as an expression
+-- on its own and resolved before the outer member-access context is known.
+-- When the type has a same-name constructor (@data Error = Error(...) | ...@),
+-- the inner resolver eagerly produces @Con (QualName Error Error) []@ for the
+-- bare name. Treat that shape as if it were @Var Error@ when it appears in
+-- qualifier position so the outer qualifier-handling cases still match.
+unwrapQualifierReceiver :: Maybe (Exp Name) -> Maybe (Exp Name)
```
axic, line 519: I have no clue about this "fix", but it does work.

rodrigogribeiro (approved):
LGTM.

PR #433, opened 2026-06-03
Renaming import fix

rodrigogribeiro:
* This PR fixes a problem in renaming functions names during imports.
* Some tests were commented to await the merge of the opcodes branch.

### DIFF
--- a/test/Cases.hs
+++ b/test/Cases.hs
@@ -82,9 +82,9 @@ dispatches :: TestTree
 dispatches =
   testGroup
     "Files for dispatch cases"
-    [ runDispatchTest "basic.solc",
-      runDispatchTest "stringid.solc",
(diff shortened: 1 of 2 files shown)


test/Cases.hs
```diff
@@ -82,9 +82,9 @@ dispatches :: TestTree
 dispatches =
   testGroup
     "Files for dispatch cases"
-    [ runDispatchTest "basic.solc",
-      runDispatchTest "stringid.solc",
-      runDispatchTest "miniERC20.solc",
+    [ -- runDispatchTest "basic.solc",
```
axic: I think this commenting out is not needed for this PR?

rodrigogribeiro: Yes, you are right. Fixed in the last commit.

mbenke (approved):

PR #434, opened 2026-06-03
Add `public` visibility modifier for contract functions

axic:
In a `contract` every function is private, unless marked as `public`. Special cases: `constructor` and `fallback` are always considered public, but the keyword is not needed to be present (if fact it can't be).

### DIFF
--- a/test/examples/dispatch/slices.solc
+++ b/test/examples/dispatch/slices.solc
@@ -8,58 +8,58 @@ import std.dispatch.{*};
 contract C {
     // --- slice_/truncate on a memory(bytes), materialized with to_bytes ---
 
-    function slice_bytes(a: memory(bytes), start: uint256) -> memory(bytes) {
+    public function slice_bytes(a: memory(bytes), start: uint256) -> memory(bytes) {
         return to_bytes(slice_(a, Typedef.rep(start)));
     }
 
-    function truncate_bytes(a: memory(bytes), end: uint256) -> memory(bytes) {
+    public function truncate_bytes(a: memory(bytes), end: uint256) -> memory(bytes) {
         return to_bytes(truncate(a, Typedef.rep(end)));
     }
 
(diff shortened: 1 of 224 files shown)


axic:
I'll rebase this once fallback is merged.

axic:
@rodrigogribeiro @mbenke this should be ready now

axic:
Rebased after comptime.

mbenke:
It seems that `public` is allowed also on top-level functions not inside a contract as well as instance methods. Is this intended? If so what is the intended meaning?

axic:
> It seems that public is allowed also on top-level functions not inside a contract as well as instance methods. Is this intended? If so what is the intended meaning?

No, that's a mistake. Will fix.

axic:
@mbenke fixed

axic:
Had to rebase due to the slices PR merge needed test updates.

test/examples/dispatch/slices.solc
```diff
@@ -8,58 +8,58 @@ import std.dispatch.{*};
...
 
-    function truncate_bytes(a: memory(bytes), end: uint256) -> memory(bytes) {
+    public function truncate_bytes(a: memory(bytes), end: uint256) -> memory(bytes) {
         return to_bytes(truncate(a, Typedef.rep(end)));
     }
 
     // --- slice_/truncate over the result of a concat ---
 
-    function slice_of_concat(a: bytes32, b: bytes32, start: uint256) -> memory(bytes) {
+    public unction slice_of_concat(a: bytes32, b: bytes32, start: uint256) -> memory(bytes) {
```
mbenke:
```suggestion
    public function slice_of_concat(a: bytes32, b: bytes32, start: uint256) -> memory(bytes) {
```

mbenke (approved):
Nice!

PR #445, opened 2026-06-09
flake: rebuild testrunner on `nix develop`

axic:
Add a shellHook that configures the CMake build directory (if missing) and runs `cmake --build build --target testrunner`. CMake's incremental build is a no-op when nothing changed, so this is cheap on warm shells but ensures the binary stays in sync with `test/testrunner/` sources.

Also add `pkgs.nlohmann_json` to the dev shell so the build can pick it up via `IGNORE_VENDORED_DEPENDENCIES=ON`, matching the path used by the Nix `checks.contests` derivation.

Set SOLCORE_SKIP_TESTRUNNER_BUILD=1 to opt out.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -154,6 +154,7 @@
             pkgs.foundry-bin
             pkgs.go-ethereum
             pkgs.jq
+            pkgs.nlohmann_json
             pkgs.solc
             evmone-lib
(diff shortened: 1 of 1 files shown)


mbenke (approved):
Looks good

PR #443, opened 2026-06-08
Improve memory helpers and add `allocate_zeroed_memory` + tests

axic:

PR #450, opened 2026-06-09
Improve the traits of the concat implementation

axic:


axic:
Would like to get #443 merged first, then I'll add tests here too.

axic:
After this a lot functions can be rewritten to be generic:
```
forall a . a:MemorySize, a:MemoryPointer => function sha256(input: a) -> bytes32 {
    let len : word = MemorySize.len(input);
    let ptr : word = MemoryPointer.ptr(input);
    let res : word;
    // We assume the [0, 32] scratch space is reserved.
    // TODO: add specific error code
    assembly {
        let ret := staticcall(gas(), 2, ptr, len, 0, 32)
        if iszero(ret) {
            revert(0, 0)
        }
        res := mload(0)
    }
    return bytes32(res);
}
```

Likewise it allows for the `bytes_slice` abstraction in the deposit PR.

mbenke (approved):
Nicely done

PR #453, opened 2026-06-10
Add cheap memory slices abstraction (with `slice_`/`truncate`)

axic:


std/std.solc
```diff
@@ -2070,6 +2073,51 @@ function zeroize_memory(ptr: word, len: word) -> () {
     }
 }
 
+// This is a very cheap abstraction over a memory area of [ptr, ptr+len)
+// No type information is preserved.
+data memory_slice = memory_slice(word, word);
```
axic: I tried a parametric version (which is started earlier in the file with `data slice(t)`, but that becomes so much more complicated and loses a ton of flexibility (like the fact this can be nested).

I very much prefer this option, but let me know. It may be also better called `memory_ref` and not a slice.

With this design we can do the following (nested to_bytes/concat/truncate/slice/hashing functions):
```
to_bytes(
  truncate(
    slice_(
      concat(truncate(a, 1)), b),
      2
    ),
    5
  )
)
```

axic: If we agree with this design, I'd like to remove `data slice(t)` to avoid naming conflict (and not have `slice_` here).

rodrigogribeiro (approved):
LGTM.

Issue #435, opened 2026-06-05
Transitive imports should not be required to be imported


axic:
If A.solc imports B.solc and B.solc imports something from C.solc, then currently A.solc must also import the given item from C.solc.

Imports in B.solc should be opaque to A.solc, otherwise this becomes rather tedious.

PR #454, opened 2026-06-10
Add msg, block, and tx global variables to std

DanielVF:
This adds stdlib support for `msg`, `block`, and `tx`, so that developers can do `msg.sender()` rather than using assembly.

  - import std.{*} works
  - import std.{msg, block, tx} works
  - msg.calldata() is used instead of msg.data(), to avoid the reserved `data` keyword.
  - block.difficulty() is intentionally absent, since it was deprecated in previous solidity versions.

Also updated the tests to remove the helper functions that were used to work around the lack of these methods.


axic:
I do wonder if the msg/tx grouping makes any importance here (compared to having `msg_caller`, `tx_origin`, etc.). The reason they were under those structs in Solidity is to reduce namespace pollution, as those were builtins. Here everything can be imported granularly, so that problem does not exists here.

It would be different to have effects or context associated with the "special" `msg` and `tx` structs, which are not available as primitives/design yet.

TLDR: I'd suggest to not have this wrapping/grouping until context or effects make it more appropriate.

std/std.solc
```diff
@@ -747,6 +752,139 @@ forall t . instance calldata(t) : Typedef(word) {
...
+forall t . class t:msg {
+  function calldata() -> calldata(bytes);
+  function sender() -> address;
+  function sig() -> bytes4;
+  function value() -> uint256;
+}
+
+forall t . default instance t:msg {
+  function calldata() -> calldata(bytes) {
+    return Typedef.abs(0) : calldata(bytes);
```
axic: Is this actually working?

std/std.solc
```diff
@@ -747,6 +752,139 @@ forall t . instance calldata(t) : Typedef(word) {
...
+  function basefee() -> uint256 {
+    let res: word;
+    assembly {
+      res := basefee()
+    }
+    return uint256(res);
+  }
+
+  function blobbasefee() -> uint256 {
+    return uint256(blobbasefee_());
```
axic: Until #435 is fixed, transitive imports are kind of clunky, hence why I'm waiting for that with #449.

std/std.solc
```diff
@@ -753,15 +753,24 @@ forall t . instance calldata(t) : Typedef(word) {
...
   function sig() -> bytes4;
   function value() -> uint256;
 }
 
 forall t . default instance t:msg {
-  function calldata() -> calldata(bytes) {
-    return Typedef.abs(0) : calldata(bytes);
+  function calldata() -> memory(bytes) {
+    let ptr: word;
+    assembly {
```
axic, line 812: You can use `allocate_zeroed_memory(round_up_to_mul_of_32(size))`.

Or for more gas efficiency:`allocate_memory()` and the use `zeroize_memory()` to clean the last (potentially) dirty bytes.

std/std.solc
```diff
@@ -753,15 +753,24 @@ forall t . instance calldata(t) : Typedef(word) {
 }
 
 forall t . class t:msg {
-  function calldata() -> calldata(bytes);
+  function calldata() -> memory(bytes);
```
axic, line 797: I'd add a comment here that this copies calldata into memory, which should change in the future once `calldata(bytes)` is operational.

std/std.solc
DanielVF: It wasn't. :D

std/std.solc
DanielVF: Will add that comment

DanielVF:
> I do wonder if the msg/tx grouping makes any importance here (compared to having `msg_caller`, `tx_origin`, etc.). The reason they were under those structs in Solidity is to reduce namespace pollution, as those were builtins. Here everything can be imported granularly, so that problem does not exists here.

I think the adoption of Core Solidity is going to live or die by how similar it feels to Solidity. People have tried to make new, better EVM languages, and adoption of all them together has been almost zero. There's danger that if Core Solidity diverges too much, people will just not use it, and stick with what they are familiar with.

So where there's not a compelling reasons to switch up syntax, I'd lean towards keeping things like classic solidity.

std/std.solc
DanielVF: Doing the latter.

DanielVF: Done.

std/std.solc
DanielVF: I switched blobbasefee to directly use assembly, like the other methods here.

DanielVF:
Will check on the test failure

mbenke:
The failing test seems to be this one: 
```
Processing: test/examples/dispatch/context.json
Compiling to Hull...
Emitting hull for contract ContextTest
Writing to output1.hull
Generating Yul...
writing output to /home/ben/work/review/build/context.yul
Compiling to bytecode...
Hex output: /home/ben/work/review/build/context.hex
context
Expected 000000000000000000000000000000000000000000000000000000000000000d but got 0000000000000000000000000000000000000000000000000000000000000012
Expected 00000000000000000000000000000000000000000000000000000000000000e1 but got 000000000000000000000000000000000000000000000000000000000000012c
```

axic:
Those are the timestamp/blocknumber constants. I think they increase with each transaction.

DanielVF:
Yeah, will just move this check into the code, against assembly.

DanielVF:
After further thought, it seems like a better idea for the tests to be deterministic inside a test file json. 

I've changed it so that block numbers and timestamps reset to known values at the start of each test json.

std/std.solc
```diff
@@ -788,6 +791,156 @@ forall t . instance calldata(t) : Typedef(word) {
     }
 }
 
+forall t . class t:msg {
+  // Currently copies calldata into memory. 
+  // Will change once calldata(bytes) is operational.
+  function calldata() -> memory(bytes);
```
axic, line 797: Btw Solidity calls this `msg.data` not `msg.calldata` 😉 

`msg.selector` would be nicer however over `msg.sig` :sigh:

std/std.solc
axic: @DanielVF meanwhile this was fixed, so the assembly helpers can be used now.

DanielVF:
I won't have time in the near future to work on this one more, so closing. Anyone can feel free to reopen and modify if they want, or start a fresh version if preferred.

PR #455, opened 2026-06-10
Fixes the pair constructor shadowing problem.

### LINKED ISSUE #438: `pair` as a keyword cannot be used in all contexts
```
contract TupleRet {
  constructor() {}

  function pair() -> (uint256, uint256) {
    return (uint256(7), uint256(11));
  }
}
```

This fails compilation:
```
Types: uint256 -> uint256 -> (uint256, uint256) and (uint256, uint256) do not unify
 - in:function pair () -> (uint256, uint256) {
   return pair(uint256.uint256(7), uint256.uint256(11));
}
```

However `pair` can be used in other contexts, see `test/examples/cases/synonym-in-function.solc`:
```
// Synonyms in function parameter and return types
type Int = word;
type Point = pair(Int, Int);

function add(a: Int, b: Int) -> Int {
    return a;
}
function makePoint(x: Int, y: Int) -> Point {
    return pair(x, y);
}
function getX(p: Point) -> Int {
    match p {
        | pair(x, _) => return x;
    }
}
function getY(p: Point) -> Int {
    match p {
        | pair(_, y) => return y;
    }
}
```

rodrigogribeiro:
Fixes #438.

### DIFF
--- a/src/Solcore/Frontend/Syntax/NameResolution.hs
+++ b/src/Solcore/Frontend/Syntax/NameResolution.hs
@@ -1029,8 +1029,11 @@ addContractName n =
   modify (\env -> env {typeEnv = Map.insert n TContract (typeEnv env)})
 
 addFunctionName :: Name -> ResolveM ()
-addFunctionName n =
-  modify (\env -> env {scopeEnv = Map.insert n TFunction (scopeEnv env)})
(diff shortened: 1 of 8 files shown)


mbenke (approved):
Looks good

PR #456, opened 2026-06-11
Fix assembly blocks in pattern matches reading/writing tuple elems

### LINKED ISSUE #451: Crash for assembly blocks using match pattern vars
```
ERROR: not implemented: Yul assembly block references variable 'v1' mapped to complex expression 'snd((1, 1))'
in: assembly {
   res := v1
}
```
for
```
contract Many {
  function main() -> word {
    let res : word;
    let foo = (1,1);
    match foo {
      | (v0, v1) => assembly { res:= v1}
    }
    return res;
   }
}
```

mbenke:
Fixes #451 and remaining TODO item in EmitHull.


rodrigogribeiro (approved):
Good catch!

PR #458, opened 2026-06-11
Enforce payable vs non-payable constructors

axic:
Constructors can now be marked `payable`. A non-payable constructor (the default) rejects any incoming value transfer at deploy time, reverting with the same NonPayableReceivedValue error (0xb5988ea3) used by the runtime method-level callvalue check. A payable constructor accepts value.

- Add `constrPayable` to the surface and resolved Constructor AST
- Parse `payable constructor(...)` (backtrack so `payable function/fallback` still parse)
- Emit a NonPayable callvalue check in the generated deployer for non-payable constructors
- Pretty-print the `payable` modifier
- Let constructor tests assert an expected outcome (status/returndata); absent "output" still means "expect success"
- Add payable_ctor / nonpayable_ctor dispatch tests and assert the existing PayableTest constructor accepts zero value

### DIFF
--- a/src/Solcore/Frontend/Pretty/SolcorePretty.hs
+++ b/src/Solcore/Frontend/Pretty/SolcorePretty.hs
@@ -161,8 +161,7 @@ instance (Pretty a) => Pretty (ContractDecl a) where
 
 instance (Pretty a) => Pretty (Constructor a) where
   ppr (Constructor ps bd payable) =
-    (if payable then text "payable" else empty)
-      <+> text "constructor"
+    (if payable then text "payable" <+> text "constructor" else text "constructor")
       <+> pprParams ps
       <+> lbrace
       $$ nest 3 (vcat (map ppr bd))
--- a/src/Solcore/Frontend/Pretty/SolcorePretty.hs
+++ b/src/Solcore/Frontend/Pretty/SolcorePretty.hs
@@ -160,8 +160,9 @@ instance (Pretty a) => Pretty (ContractDecl a) where
     ppr c
 
 instance (Pretty a) => Pretty (Constructor a) where
-  ppr (Constructor ps bd) =
-    text "constructor"
+  ppr (Constructor ps bd payable) =
+    (if payable then text "payable" else empty)
+      <+> text "constructor"
       <+> pprParams ps
       <+> lbrace
       $$ nest 3 (vcat (map ppr bd))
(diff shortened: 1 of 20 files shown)


axic:
Rebased after comptime.

test/testrunner/testrunner.cpp
```diff
@@ -141,14 +141,58 @@ int main(int argc, char** argv)
...
+					else if (expectedStatus == "success")
+					{
+						if (!status)
+						{
+							std::cerr << "Expected creation success but got failure" << std::endl;
+							resultRecorder.record(filename, "Expected constructor status success but got failure.", "failure", expectedStatus, gasUsed, gasUsedForDeposit);
+							hasTestFailure = true;
+							continue;
+						}
+					}
```
mbenke, line 171: what if expected status is not one of `["failure","success"]` (e.g. typo)?

src/Solcore/Frontend/Pretty/TreePretty.hs
```diff
@@ -136,8 +136,9 @@ instance Pretty ContractDecl where
     ppr c
 
 instance Pretty Constructor where
-  ppr (Constructor ps bd) =
-    text "constructor"
+  ppr (Constructor ps bd payable) =
+    (if payable then text "payable" else empty)
+      <+> text "constructor"
```
mbenke: This introduces an extra space if not payable


```suggestion
    if payable then text "payable" <+> text "constructor" else text "constructor"
```

src/Solcore/Frontend/Pretty/SolcorePretty.hs
```diff
@@ -160,8 +160,9 @@ instance (Pretty a) => Pretty (ContractDecl a) where
     ppr c
 
 instance (Pretty a) => Pretty (Constructor a) where
-  ppr (Constructor ps bd) =
-    text "constructor"
+  ppr (Constructor ps bd payable) =
+    (if payable then text "payable" else empty)
+      <+> text "constructor"
```
mbenke: This introduces an extra space if not payable


```suggestion
    if payable then text "payable" <+> text "constructor" else text "constructor"
```

mbenke (approved):

test/testrunner/testrunner.cpp
axic: This is mostly a code duplication of the non-constructor part below.  I can add an error in both cases.

src/Solcore/Frontend/Pretty/TreePretty.hs
axic: Is it better two have two tokens, than one `text "payable constructor"` ?

axic:
@mbenke pushed the fixes

PR #461, opened 2026-06-12
Restrict `payable` to contract functions/fallback

axic:
`payable` was accepted on any function signature, including top-level functions, class signatures, and instance methods. Thread an `allowPayable` flag through signatureP/funDefAfterPrefix so it is only accepted on a function or fallback inside a `contract`; elsewhere it fails with a clear message.

`constructor` and `fallback` are only wired into contract bodies, but writing them at the top level produced a confusing generic parse error. Add an explicit top-level guard that rejects them with a clear "only inside a contract" message.

Add expect-failure fixtures for top-level payable function, top-level constructor, and top-level fallback.


src/Solcore/Frontend/Parser/Decl.hs
```diff
@@ -350,15 +363,28 @@ topDeclP =
...
               ]
         )
     ]
 
+-- | @constructor@ and @fallback@ declarations are only meaningful inside a
+-- @contract@. Catch them at the top level so we report a clear error instead
+-- of a confusing generic parse failure. Each branch commits (consumes the
+-- keyword) before failing, so the surrounding 'choice' does not fall through
+-- to the function/class/instance parser.
+contractOnlyDeclP :: Parser TopDecl
```
axic, line 412: These just provide a nicer error message, not strictly neccesary.

Before:
```
<input>:8:1:
  |
8 | constructor() {}
  | ^
unexpected 'c'
expecting "class", "contract", "data", "default", "export", "forall", "function", "import", "instance", "payable", "pragma", "type", or end of input
```

After:
```
<input>:8:12:
  |
8 | constructor() -> () {}
  |            ^
a `constructor` may only be declared inside a contract
```

mbenke (approved):
Well done

PR #463, opened 2026-06-14
Update primitives/opcodes to Osaka (and fix some bugs)

axic:
This
- sorts the primitives to match the opcode order (see https://evm.codes),
- removes duplicate `difficulty` (of `prevrandao`, as it was renamed several hardforks back),
- removes `pc` as it is not a Yul builtin,
- adds the missing opcodes as per Osaka hardfork (@d-xo mentioned a while back the current goal is to match latest hardfork, then we can decide if/how to support older ones),
- and adds missing specific Yul builtins,
- fix wrong definition of `extcodecopy` in Hull,
- remove `pc/`jump`/`jumpi` from Hull,
- sync the primitives with Hull.


mbenke (commented):
In `std/opcodes.solc`, it may be worthwhile to add a comment along the lines

```
  // Generated by scripts/gen-std-opcodes.py. Run the script to regenerate.
```

mbenke (approved):

axic:
@mbenke added

PR #466, opened 2026-06-15
Add a compiler pass for automatic generation of Generic instances for algebraic datatypes

rodrigogribeiro:
* Add a compiler pass for generating Generic instance for datatypes.
* Add a pragma to disable Generic instance generation for datatypes.


test/examples/dispatch/generic_sum.solc
```diff
@@ -9,45 +9,45 @@ pragma no-bounded-variable-condition;
...
         }
     }
-    function to(r : Sum((), uint256)) -> Option(uint256) {
+    function to(r : sum((), uint256)) -> Option(uint256) {
         match r {
-        | Sum.Inl(_) => return Option.None;
-        | Sum.Inr(v) => return Option.Some(v);
+        | inl(_) => return Option.None;
+        | inr(v) => return Option.Some(v);
         }
```
mbenke, line 26: This instance should be derived, right? And similarly in `generic_product`?
Or perhaps there should be two sets of tests: one for explicit instance, and one for derived?

rodrigogribeiro: Fixed and added tests for the pragma that disable Generic instance generation.

test/examples/cases/user-op-lambda.solc
```diff
@@ -0,0 +1,20 @@
+import std.{*};
+pragma no-patterson-condition ;
+pragma no-coverage-condition ;
+pragma no-bounded-variable-condition ;
+
+infixl 70 (^^) => pow;
```
mbenke, line 6: I don't think the operators have been merged yet:

```
test/examples/cases/user-op-lambda.solc
<input>:6:1:
  |
6 | infixl 70 (^^) => pow;
  | ^
unexpected 'i'
expecting "class", "constructor", "contract", "data", "default", "export", "fallback", "forall", "function", "import", "instance", "payable", "pragma", "public", "type", or end of input
```

mbenke (approved):
Nice work!

PR #467, opened 2026-06-15
Fix transitive import bug

### LINKED ISSUE #435: Transitive imports should not be required to be imported

Y-Nak:
Resolves #435


rodrigogribeiro (approved):
Great work!

PR #469, opened 2026-06-16
Implement storage arrays

axic:


mbenke (changes requested):
My main concern here is the UFCS - commits (e.g. 10823085) suggest that it was supposed to be moved to a separate PR but was not, but tests have been removed.

It might be cleaner to move UFCS to separate PR, but probably less work to add a description and tests for it (mainly that it does not conflict with other uses of dot syntax such as qualified function calls and structure members).

rodrigogribeiro:
> My main concern here is the UFCS - commits (e.g. [1082308](https://github.com/argotorg/solcore/commit/10823085054c7b184649b92ea62cbfff905859f7)) suggest that it was supposed to be moved to a separate PR but was not, but tests have been removed.
> 
> It might be cleaner to move UFCS to separate PR, but probably less work to add a description and tests for it (mainly that it does not conflict with other uses of dot syntax such as qualified function calls and structure members).

Added some comments documenting this and added a test case.

mbenke (approved):

PR #470, opened 2026-06-16
Rewrite sha256/ripemd160/ecrecover without assembly and add `Error`

axic:


std/std.solc
```diff
@@ -2184,31 +2186,19 @@ forall a . a:MemorySize, a:MemoryPointer => function sha256(input: a) -> bytes32
...
-    assembly {
-        let ret := staticcall(gas(), 2, ptr, len, 0, 32)
-        if iszero(ret) {
-            revert(0, 0)
-        }
-        res := mload(0)
-    }
-    return bytes32(res);
+    let ret = staticcall(gas(), 2, ptr, len, 0, 32);
+    require(ret != 0, Error(0x68c071bb)); // SHA256CallFailed()
```
axic, line 2134: Now it would be nice having gas comparison tests, as I do wonder what is the overhead of `require`. Ideally nothing.

mbenke (approved):

PR #471, opened 2026-06-16
Remove obsolete mload/sload helpers

axic:
The `std.opcodes` helpers can be used where, otherwise for storage the `StorageType` class is used everywhere.


rodrigogribeiro (approved):
LGTM.

PR #475, opened 2026-06-17
Emit JSON ABI files for contracts

mbenke:
Add ABI generation to contract dispatch: contractAbiJson renders a contract's public functions, constructor and fallback as a JSON array in the layout used by solc, and writeContractAbis writes one <ContractName>.abi per contract.

Gate emission behind a new --abi flag, and add -o/--output-dir to control where generated files (.abi and .hull) are written, creating the directory if needed.

Update runsol.sh to pass -o instead of moving output*.hull into build/.

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/runsol.sh
+++ b/runsol.sh
@@ -191,14 +191,10 @@ fi
 
 # Execute compilation pipeline
 echo "Compiling to hull..."
-if ! cabal run sol-core -- -f "$file"; then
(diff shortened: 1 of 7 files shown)


rodrigogribeiro (approved):
LGTM.

PR #478, opened 2026-06-19
Fix calldatasize check for fallback

axic:


std/dispatch.solc
```diff
@@ -160,7 +161,7 @@ forall args rets fn
     assembly {
       cdsz := calldatasize()
     }
-    if ((cdsz - 4) < hdsz) {
+    if (cdsz < (hdsz + 4)) {
```
axic, line 164: The original code her resulted in underflow (wraparound), which "kind of" made the flow work.

The problem is that if this gets fixed here, then the `do_exec()` chain ends up in both cases (cdsz < 4 and cdsz >= 4) in the variant where `cdsz < 4` is checked (the check-decode-exec-encode variant).

I fixed this with specializing the fallback case to not reuse that code.

This then showed yet another issue: without the implicit stop/return(0,0), we end up with `main` function's return:
```
mstore(0, _mainresult)
return(0, 32)
```

rodrigogribeiro (approved):
LGTM.

PR #479, opened 2026-06-19
materialize string literals into memory(string)

mbenke:
`string` is a comptime-only type with no runtime representation.  String literals and comptime concatenations are now polymorphic via a primitive `Str` class and are materialized into `memory(string)` at runtime sites, mirroring the comptime `integer`/`Int.fromInteger` design.

- StrLiteralDesugar: untyped pre-typecheck pass wrapping every string literal and every `concatLit(...)` in `Str.fromString(...)`, resolved by the type checker per use site (identity at `string`, materialization at `memory(string)`).  Wrapping `concatLit` makes it effectively result-polymorphic, so `concatLit(...)` (incl. nested) works at a `memory(string)` site without a manual `Str.fromString`.
- Primitives/TcEnv/NameResolution: `Str` class with bodyless `string` and `memory(string)` instances; `memStringFromLit` backend marker.
- Specialise: `Str.fromString` resolves by result type to identity or `memStringFromLit`.
- MastEval: `Str.fromString` identity fold; `memStringFromLit` kept impure.
- EmitHull: intercept `memStringFromLit [StrLit]` -> per-literal, content- deduped `__strlit_<n>` allocator (alloc + length + char words); clean guards when a comptime-only type reaches emission.
- TcStmt: generalize markIntegerComptime -> markComptimeOnly (integer+string).
- ComptimeCheck: classify comptime-only-typed params/returns as comptime, so `concatLit` and string functions are implicitly comptime without cascading the annotation to callers.

Tests: comptime materialization (basic, concat, nested, dedup, negative) and an end-to-end dispatch case returning "Hello, world!" verified via evmone. Documented in doc/comptime-string.md.


rodrigogribeiro (approved):
Great work!

PR #482, opened 2026-06-20
Implement `erc7201`

axic:
This follows Solidity: https://docs.soliditylang.org/en/v0.8.35/cheatsheet.html#mathematical-and-cryptographic-functions

While this works, it will be some feat making this `comptime` (but a good challenge at the same time).

The test case is the canonical example from the standard: https://eips.ethereum.org/EIPS/eip-7201


std/std.solc
```diff
@@ -67,6 +67,7 @@ export {
   concatLit,
   ecrecover,
   empty(*),
+  erc7201,
   eqWord,
```
mbenke:
```suggestion
  eqWord,
  erc7201,
```

mbenke (approved):

PR #485, opened 2026-06-23
Add bitwise and modulo syntactic sugar

axic:
And implement it for `word`.

The first commit also enables MastEval for bandWord, borWord, bxorWord.


std/std.solc
```diff
@@ -487,6 +506,39 @@ function bshrWord(x: word, y: word) -> word {
...
+// 2-argument bitwise OR.
+// Intended to be folded by MastEval when both arguments are
+// statically known word values.
+function orWord(a : word, b : word) -> word {
+    let res : word;
+    assembly {
+        res := or(a, b)
+    }
+    return res;
+}
```
mbenke: These seem to duplicate `bandWord/borWord/bxorWord`

mbenke (changes requested):
Alas, this PR breaks the parser - `|` as an infiox operator conflicts with `|` as match case separator: 

```
ben@trawa:~/work/review$ grep -v // tmp/parse485.solc

data Opt = None | Some(());

function f(o: Opt) -> () {
  match o {
    | .Some(x) => x
    | .None => ()
  }
}
ben@trawa:~/work/review$ cabal run -- sol-core -f tmp/parse485.solc
<input>:18:19:
   |
18 |     | .Some(x) => x
   |                   ^
unexpected 'x'
expecting '|' or '}'
```

axic:
@mbenke sorry forgot to mark draft, this wasn't fully ready.

std/std.solc
axic: Yes, this will replace it.

std/std.solc
```diff
@@ -473,6 +488,24 @@ instance word:Mul {
...
+        return bandWord(l, r);
+    }
+}
+
+instance word:BitOr {
+    function bor(l: word, r: word) -> word {
+        return borWord(l, r);
+    }
+}
+
```
mbenke: why no instances for uint256? For one it makes an e2e test (which is conspicuously missing) more difficult.

mbenke (changes requested):
Add an e2e (dispatch) test and perhaps uint256 instances

axic:
@mbenke done

std/std.solc
axic: Added.

mbenke (approved):

PR #486, opened 2026-06-23
Fix returnCount arity for sum-returning HULL functions

axic:
returnCount in Language/Hull/TcEnv.hs collapsed every non-TFun/TPair type to a single word slot, so a sum-returning function called from a user asm block passed HULL's YAssign/YLet arity check (1 == 1). yule's sizeOf, however, lowers a TSum return as `1 + max (sizeOf t1) (sizeOf t2)` words, emitting a multi-word Yul return into a single-slot LHS that solc later rejects.

Make returnCount mirror yule's sizeOf exactly by adding explicit TSum, TSumN and TBool cases (and an explicit TFun error), so the asm arity check agrees with the eventual Yul lowering.

Add a Hull type-checker test (11-err-asm-sum-return.hull) that assigns a `word + word` returning function into a single word in an assembly block; it now correctly fails the return-count check.

### DIFF
--- a/src/Language/Hull/TcEnv.hs
+++ b/src/Language/Hull/TcEnv.hs
@@ -27,12 +27,19 @@ emptyHullTcEnv :: HullTcEnv
 emptyHullTcEnv = HullTcEnv Map.empty hullBuiltins Nothing
 
 -- Number of word-sized slots a return type occupies.
+-- This MUST mirror yule's @sizeOf@ (yule/Translate.hs): sum types lower to a
+-- tag word plus the widest variant payload, so they occupy more than one slot.
+-- Collapsing them to 1 here lets a sum-returning function be assigned into a
+-- single-word asm LHS that yule then lowers to a multi-word Yul return.
 returnCount :: Type -> Int
 returnCount TUnit = 0
 returnCount TWord = 1
+returnCount TBool = 1
 returnCount (TPair a b) = returnCount a + returnCount b
+returnCount (TSum a b) = 1 + max (returnCount a) (returnCount b)
+returnCount (TSumN ts) = 1 + maximum (map returnCount ts)
 returnCount (TNamed _ t) = returnCount t
-returnCount _ = 1
+returnCount (TFun _ _) = error "returnCount: function type has no runtime representation"
 
 -- Return type for exactly n word-sized return slots.
 nReturns :: Int -> Type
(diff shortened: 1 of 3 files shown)


src/Language/Hull/TcEnv.hs
```diff
@@ -27,12 +27,19 @@ emptyHullTcEnv :: HullTcEnv
...
+-- tag word plus the widest variant payload, so they occupy more than one slot.
+-- Collapsing them to 1 here lets a sum-returning function be assigned into a
+-- single-word asm LHS that yule then lowers to a multi-word Yul return.
 returnCount :: Type -> Int
 returnCount TUnit = 0
 returnCount TWord = 1
+returnCount TBool = 1
 returnCount (TPair a b) = returnCount a + returnCount b
+returnCount (TSum a b) = 1 + max (returnCount a) (returnCount b)
+returnCount (TSumN ts) = 1 + maximum (map returnCount ts)
```
axic, line 40: Not sure if these are correct?

I think the error below should be enough?

mbenke: Compressed sums can change the outcome and with the current pipeline we have no way of knowing whether the compression will be enabled or not.

See `test/examples/spec/037dwarves.solc` for a bigger sum example (compile with `-s -g --pe-fuel 0` to see sums in the Hull output).

There is also the comment 

```
-- Collapsing them to 1 here lets a sum-returning function be assigned into a
-- single-word asm LHS that yule then lowers to a multi-word Yul return.
```

(from the original code, but unchanged in this PR) which I do not understand, @rodrigogribeiro, can you help?

mbenke: Having thought about this for a while, I think a proper fix might be to remove unoptimised sums from yule and change the `returnCount` formula or `TNamed` appropriately.

rodrigogribeiro: I my view, the `returnCount` implementation for `TSum` and `TSumN` makes sense. Is there any problem that I was not able to perceive?

mbenke: I was worried that compressing sums may break this arithmetic, but compression actually runs before compression, so this is fine. Sorry.

mbenke (approved):

rodrigogribeiro (approved):
I believe that this change will fix a problem that I've found on implementing generic deriving. Nice catch!

PR #487, opened 2026-06-23
Constrain Yul assignment LHS to `word`

axic:


test/examples/cases/yul-asm-for-body.solc
```diff
@@ -1,7 +1,7 @@
 import std.{*};
 
 function yul_asm_for_body() -> () {
-    let result : uint256 = uint256(0);
+    let result : word = 0;
```
axic, line 4: These three test cases were added in #363 -- makes me wonder if that "fix" wasn't fully correct, and it was eliminated to begin with as the types didn't even match.

rodrigogribeiro (approved):
LGTM

PR #489, opened 2026-06-23
Add optional `gasUsed` assertion to testrunner output expectations

axic:
Added it to three test cases, which use quite a few features of the stdlib.

Pro: Maybe this is too early, but would be nice to track effects. 
Con: Very likely many (if not all) stdlib PRs will have to modify these three values.


axic:
Please note: before merging this needs to be updated as likely other changes affected the gasUsed already.

rodrigogribeiro (approved):
LGTM

axic:
@rodrigogribeiro rebased with latest numbers.

PR #493, opened 2026-06-24
Fix tcYulStmt dropping uninitialized `let` and crashing on bool literals

### LINKED ISSUE #491: Let-statements without assignment are ignored in assembly
```
    public function f() -> () {
        let k: word;
        assembly { 
            let x
            k := x
        }
    }
```
->
```
Undefined name: x
 - in:x
 - in:public function f () -> () {
   let k : word ;
   assembly {
      let x
      k := x
   }
}
```

### LINKED ISSUE #492: `true`/`false` in Yul panics
```
    public function f() -> () {
        assembly {
            let x := true
        }
    }
```
->
```
CallStack (from HasCallStack):
  error, called at src/Solcore/Frontend/TypeInference/TcStmt.hs:1963:28 in sol-core-0.0.0.0-inplace:Solcore.Frontend.TypeInference.TcStmt
  notImplemented, called at src/Solcore/Frontend/TypeInference/TcStmt.hs:1776:14 in sol-core-0.0.0.0-inplace:Solcore.Frontend.TypeInference.TcStmt
```

The `true`/`false` literals in Yul exists and they correspond to 1/0.

axic:
Fixes #491
Fixes #492

`tcYulStmt` only matched `YLet ns (Just e)`; uninitialized `let x` (`YLet ns Nothing`) fell into the `tcYulStmt _ = pure ([], unit)` catch-all, so `x` never entered the env and later assignments/reads of it bypassed the LHS arity/type check. The same catch-all also silently swallowed `YFun` (Yul function bodies never typed), `YBreak`, `YContinue`, `YLeave`, and `YComment`. Additionally a literal `true` / `false` in a Yul block hit `tcYLit ... = notImplemented`, crashing the compiler.

- Handle `YLet` with and without an initializer; both introduce `word` bindings into the env.
- Type-check `YFun` bodies with parameters and named returns bound, and register the function name so calls (and recursion) resolve.
- Make `YBreak`/`YContinue`/`YLeave`/`YComment` explicit no-ops and drop the catch-all so the match is exhaustive.
- Treat Yul `true`/`false` as `word` literals (no Yul boolean type).
- Add regression tests asm-let-uninit.solc and asm-let-bool-lit.solc.

### DIFF
--- a/test/examples/dispatch/assembly.solc
+++ b/test/examples/dispatch/assembly.solc
@@ -1,0 +1,19 @@
+import std.{*};
+import std.dispatch.{*};
+
+contract C {
+    constructor() {}
+
+    // Exercises a Yul block that declares an uninitialized `let y`, assigns the
+    // boolean literal `true` to it, and writes it back to the surrounding
+    // `word` local `x`. `true` is the word `1`, so this returns uint256(1).
+    public function asmBool() -> uint256 {
+        let x : word;
+        assembly {
+          let y
+          y := true
+          x := y
+        }
+        return uint256(x);
+    }
+}
--- a/run_contests.sh
+++ b/run_contests.sh
@@ -6,6 +6,7 @@ root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$root_dir"
 
 bash ./contest.sh test/examples/dispatch/basic.json
+bash ./contest.sh test/examples/dispatch/assembly.json
 bash ./contest.sh test/examples/dispatch/neg.json
 bash ./contest.sh test/examples/dispatch/miniERC20.json
 bash ./contest.sh test/examples/dispatch/Revert.json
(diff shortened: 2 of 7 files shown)


test/examples/dispatch/assembly.solc
```diff
@@ -0,0 +1,19 @@
...
+    constructor() {}
+
+    // Exercises a Yul block that declares an uninitialized `let y`, assigns the
+    // boolean literal `true` to it, and writes it back to the surrounding
+    // `word` local `x`. `true` is the word `1`, so this returns uint256(1).
+    public function asmBool() -> uint256 {
+        let x : word;
+        assembly {
+          let y
+          y := true
```
axic, line 14: We can argue whether we want to support true/false here, especially as it not interchangeable with solcore's.

If not, instead of a panic it should be a parser error -- "true/false literals are not accepted".

mbenke (approved):

rodrigogribeiro (approved):
Great work!

PR #494, opened 2026-06-24
Guard constructor argSize against codesize underflow

axic:
In copy_arguments_for_constructor, argSize = sub(codesize(), programSize) underflows to ~2^256 when init-code is truncated (codesize() < programSize), corrupting the free-memory pointer and codecopy length. Add an explicit lt(codesize(), programSize) check that reverts before the subtraction.

### DIFF
--- a/src/Solcore/Desugarer/ContractDispatch.hs
+++ b/src/Solcore/Desugarer/ContractDispatch.hs
@@ -169,13 +169,8 @@ transformConstructor contractName cons
             Let False "memoryDataOffset" (Just word) Nothing,
             Asm
               [yulBlock|{
+                 // NOTE: we ensure no truncation in startBody below
                  let programSize := datasize(`deployer`)
-                 // Guard against argSize underflow: a truncated init-code
-                 // (codesize() < programSize) would make sub() wrap to ~2^256,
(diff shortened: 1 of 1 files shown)


src/Solcore/Desugarer/ContractDispatch.hs
```diff
@@ -170,6 +170,11 @@ transformConstructor contractName cons
             Asm
               [yulBlock|{
                  let programSize := datasize(`deployer`)
+                 // Guard against argSize underflow: a truncated init-code
+                 // (codesize() < programSize) would make sub() wrap to ~2^256,
+                 // corrupting the free-memory pointer and codecopy length.
+                 // TODO: use require with proper error
+                 if lt(codesize(), programSize) { revert(0, 0) }
```
axic: In the case a truncated bytecode is sent, the `codeCopy` below would try to write into infinitely large memory pointer, effectively causing and out-of-gas error.

There's no easy way to test this, as testrunner sends the entire code.

axic: Does make me wonder, this check probably should go into `startBody` before `copy_arguments_for_constructor`, so it is run even if no arguments.

rodrigogribeiro (approved):
LGTM

PR #495, opened 2026-06-24
Add a sanity check to `*Lit()` functions

axic:
There is a proper runtime error, and not a silent bug, in case the constant folding somehow skips them.


rodrigogribeiro (approved):
LGTM

PR #496, opened 2026-06-24
Add `std.raw_call`

axic:
This follows `<address>.call()` from Solidity (see https://docs.soliditylang.org/en/v0.8.35/cheatsheet.html#members-of-address).  We could wrap this into something like `Address.call()`, but for now I'd prefer this simple version.


std/std.solc
```diff
@@ -2109,3 +2110,24 @@ function erc7201(id: memory(bytes)) -> bytes32 {
...
+        MemoryPointer.ptr(payload),
+        MemorySize.len(payload),
+        0,
+        0
+    );
+    let retSize = returndatasize();
+    let retData = allocate_memory(32 + retSize);
+    mstore(retData, retSize);
+    // TODO: use returndatacopy(retData + 32, 0, retSize);, but it is a parser error
+    // See https://github.com/argotorg/solcore/issues/497
```
mbenke, line 2128: can be changed as soon as #504 is merged

mbenke (approved):

rodrigogribeiro (approved):
LGTM

PR #501, opened 2026-06-24
Add storage support for `bytes` fields

axic:
Mirror the existing string/memory(string) storage support for bytes:
- StorageSize instance for memory(bytes)
- CanStore instance for storage(memory(bytes)) that copies to/from storage using the existing storeBytesFromMemory / loadBytesFromStorage helpers (bytes and string share an identical storage layout)

This lets a contract declare a `memory(bytes)` field and assign/read it, e.g.

  contract C {
    content: bytes;
    public function set(value: memory(bytes)) -> () { content = value; }
    public function get() -> memory(bytes) { return content; }
  }

Adds a storage dispatch test (compile + EVM round-trip for both short in-place and long out-of-place byte arrays).


mbenke (approved):

rodrigogribeiro (approved):
LGTM.

PR #504, opened 2026-06-25
Make parser recognise identifiers that start with a keyword

### LINKED ISSUE #497: Parser for `return` is too greedy
```
import std.opcodes.returndatacopy;
function f() -> () {
    returndatacopy(0, 0, 32);
}
```
->
```
<input>:2116:11:
     |
2116 |     returndatacopy(0, 0, 32);
     |           ^
unexpected 'd'

Error: sol-core compilation failed
```

mbenke:
Fixes #497

### DIFF
--- a/src/Solcore/Frontend/Lexer/SolcoreLexer.hs
+++ b/src/Solcore/Frontend/Lexer/SolcoreLexer.hs
@@ -32,8 +32,11 @@ lexeme = L.lexeme sc
 symbol :: String -> Parser String
 symbol = L.symbol sc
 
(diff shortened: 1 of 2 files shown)


axic:
Also found this:
```
<input>:1797:21:
     |
1797 |             let data = and_(value, mask);
     |                     ^
reserved word used as identifier: data
```

I think this was added as workaround for the same problem? Or the parser is not context-aware? Or is it to avoid confusion for it being a type declaration?

rodrigogribeiro (approved):
LGTM

PR #506, opened 2026-06-27
Clean ABI output when storing string/bytes

axic:


std/std.solc
```diff
@@ -1192,7 +1192,9 @@ function encodeIntoFromBytesLike(srcPtr:word, basePtr:word, offset:word, tail:wo
     let total = length + 32;
     mstore(basePtr + offset, tail - basePtr);
     mcopy(tail, srcPtr, total);
-    return tail + round_up_to_mul_of_32(total);
+    let rounded = round_up_to_mul_of_32(total);
+    zeroize_memory(tail + total, rounded - total);
+    return tail + rounded;
```
axic, line 1196: We return a 32-byte aligned pointer, but we didn't make sure the memory area is zero.

In the current codebase the memory area *is* zero as nothing is deallocated, but this matches the logic of Solidity and allows for deallocating memory (or rather having scratch memory which is discarded, which does happen in Solidity).

rodrigogribeiro (approved):
LGTM.

PR #508, opened 2026-06-29
Fail loudly when ABI emitter hits a parameterized type

axic:
The catch-all `abiTypeOf (TyCon n _)` matched any type constructor regardless of arity, discarding its type arguments. A public function with a parameterized parameter type (e.g. `mapping(word, word)` or a custom generic) emitted a bare, invalid ABI string like `"type":"mapping"`, which ABI consumers (etherscan, ethers, web3py) misparse. This also shadowed the loud-error fallthrough below it.

Restrict the passthrough to nullary type constructors (which is what all legitimate ABI value types — address, bool, bytes32, uint256, bytes, string, word — are), so parameterized types fall through to the existing "cannot represent type in ABI" error instead of silently producing malformed output. Selectors are unaffected.

Add a ContractAbiTests case asserting a mapping-typed parameter fails loudly.


rodrigogribeiro (approved):
LGTM.

PR #511, opened 2026-06-29
Fixes in the specializer for dealing with matches

rodrigogribeiro:
Consider the following contrived example:

```
forall f g . f:Total, g:Total => instance sum(f, g) : Total {
  total(x : sum(f,g), y : sum(f,g)) =
    match x {
    | inl(xa) => match y { inl(ya) => Total.total(xa, ya)   // recurses at f
                         | inr(yb) => 0 }
    | inr(xb) => match y { inl(ya) => 0
                         | inr(yb) => Total.total(xb, yb) }  // recurses at g
    }
}
```

Specializing the call at `sum((word,word), word)` gives `f = pair(word,word)` and `g = word`. Note
that the two outer branches recurse at **different types**: the `inl` branch at `f` (a pair), the
`inr` branch at `g` (a word).

1. Specializing the `inl` branch, the recursive call `Total.total(xa, ya)` is resolved at `f`. That
   triggers `extSpSubst`, pinning down that branch's metavariables. In particular, a payload
   metavariable that is unused/free in this branch ends up identified, and the net effect is an
   identification `f ~ g` in the global accumulator.
2. The loop proceeds to the `inr` branch **without clearing `spSubst`**. Recomputing the scrutinee
   of that branch's `match y`, `atCurrentSubst` applies the inherited binding to `y : sum(f,g)` and
   collapses it to `sum(g,g)`, that is, `sum(word,word)`.
3. The emitted `.hull` then annotates that inner `match` with `sum(word,word)`, while the value that
   actually flows in is `sum(pair(word,word), word)`.

The frontend does not notice (it type-checked before specialization). But `yule` revalidates types
structurally and aborts:

```
Type mismatch
  expected: sum(word, word)              -- sum(g,g), collapsed
  actual:   sum(pair(word,word), word)   -- sum(f,g), correct
```

* Why only "sum of product" breaks

If the two sides of the sum have the **same shape** (for example `sum(word, word)`, an enum, and so
on), the `f ~ g` collapse is a no-op: `word` becomes `word`, the corrupted type coincides with the
correct one, and `yule` sees no discrepancy. The bug only becomes **visible** when `f` and `g` have
different shapes (pair vs word), which is exactly the "constructor with more than one field" case.
That is why enums and uniform sums always passed, and only sum-of-product failed.

* How the fix works

```haskell
exps' <- specScruts exps
saved <- getSpSubst               -- snapshot AFTER the scrutinees
alts' <- forM alts $ \alt -> do
  putSpSubst saved                -- each alt starts from the same clean state
  specAlt alt
putSpSubst saved                  -- nothing leaks past the match
return $ Match exps' alts'
```

The idea is to give each alternative the **substitution scope** it should have:

- `saved` is captured *after* `specScruts`, so it includes everything the call context and the
  scrutinees legitimately determined (`f = pair`, `g = word`). That is what all alternatives share.
- `putSpSubst saved` before each alternative ensures it is specialized under exactly that shared
  context, **without** the bindings a sibling discovered internally. So each branch's `match y`
  scrutinee is seen as the intact `sum(f,g)`.
- `putSpSubst saved` after the loop prevents a branch-local binding from escaping into statements
  that come after the `match`.

### DIFF
--- a/run_contests.sh
+++ b/run_contests.sh
@@ -24,5 +24,6 @@ bash ./contest.sh test/examples/dispatch/storage.json
 bash ./contest.sh test/examples/dispatch/generic_sum.json
 bash ./contest.sh test/examples/dispatch/generic_product.json
 bash ./contest.sh test/examples/dispatch/sum_wide_product.json
+bash ./contest.sh test/examples/dispatch/specialise_sum_of_product.json
 bash ./contest.sh test/examples/dispatch/forloops.json
 bash ./contest.sh test/examples/dispatch/weth9.json
--- a/src/Solcore/Backend/Specialise.hs
+++ b/src/Solcore/Backend/Specialise.hs
@@ -746,7 +746,11 @@ specMatch exps alts = do
   -- subst <- getSpSubst
   -- debug ["> specMatch, scrutinee: ", pretty exps, " @ ", pretty subst]
   exps' <- specScruts exps
-  alts' <- forM alts specAlt
+  saved <- getSpSubst
+  alts' <- forM alts $ \alt -> do
+    putSpSubst saved
+    specAlt alt
+  putSpSubst saved
   -- debug ["< specMatch, alts': ", show alts']
   return $ Match exps' alts'
   where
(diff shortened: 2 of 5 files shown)


mbenke (approved):
Good catch!

PR #515, opened 2026-07-01
Move yule backend logic into the sol-core library

mbenke:
The Yul backend lived in the `yule` executable's own top-level modules. Move them into the library under a proper namespace and extract the Yul-object assembly logic out of `yule/Main.hs`, so the hull -> Yul step is reusable from the library (CLI, tests and other tools):

  Compress  -> Language.Hull.Compress
  Builtins  -> Language.Yul.Builtins
  Locus, Options, TM, Translate, and the new assembly module
            -> Language.Hull.ToYul.*

- The yule executable now contains only Main.hs; it imports wrapInObject from Language.Hull.ToYul.Assemble and the translation modules by their new names.
- Translate: treat SComment as a no-op in genStmt. The hull parser skips block comments, so the CLI never saw them, but emitHull emits comment statements and they hit genStmt's error catch-all when a hull object is translated in-process (without a serialize/parse round-trip).

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/sol-core.cabal
+++ b/sol-core.cabal
@@ -112,12 +112,21 @@ library
       Solcore.Pipeline.SolcorePipeline
       Solcore.Primitives.Primitives
       Language.Hull
+      Language.Hull.Compress
       Language.Hull.Parser
       Language.Hull.TcEnv
       Language.Hull.TcMonad
       Language.Hull.TypeCheck
       Language.Hull.Types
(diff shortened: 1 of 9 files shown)


Y-Nak (approved):
LGTM

rodrigogribeiro (approved):
LGTM.

PR #516, opened 2026-07-01
Add instance uint256:Int and fix bugs that prevented it

mbenke:
`uint256` did not have `Int` instance required to automatically convert literals.
With these changes assignments like `counter=42` do not need explicit `uint256`


Y-Nak (approved):
LGTM

rodrigogribeiro (approved):
Great work!

PR #521, opened 2026-07-04
Typechecker performance improvements

rodrigogribeiro:
This PR makes some changes on type inference to improve performance.

- Single type-inference pass: merged the two full inference passes into one.
- SYB span-stopping: A lot of typing time is spent by SYB traversals. The improvement is to make them not to check the AST Span information using new combinators in `Traversal.hs`
- `Subst` as `Data.Map`: `apply` is now `O(log n)` instead of a linear association-list scan.
- `ordNub` in `fv`/`mv`/`bv`: replaced `O(n^2)` `Data.List.union` accumulation.
- Minor optimizations: removed dead `nameSupply` mutation in `freshInst`; `askInstEnv`
  alpha-renames only the queried class's instances, not the whole table.

Times in seconds (minimum of several warm runs on my machine). `main` is the original
two-pass baseline; `optimized` is this branch. `tc` = "Typecheck modules" phase; `wall` = total
compile time.

| Program     | tc: main | tc: optimized | tc speed-up | wall: main | wall: optimized | wall speed-up |
|-------------|---------:|--------------:|:-----------:|-----------:|----------------:|:-------------:|
| storage     |   4.16   |     0.74      |   5.62x     |    5.21    |      1.04       |    5.00x      |
| generic_sum |   6.01   |     1.01      |   5.95x     |    7.43    |      1.39       |    5.33x      |
| weth9       |   4.86   |     0.86      |   5.65x     |    6.09    |      1.22       |    5.00x      |
| forloops    |   4.53   |     0.82      |   5.52x     |    5.68    |      1.16       |    4.88x      |
| miniERC20   |   5.30   |     0.97      |   5.46x     |    6.50    |      1.30       |    5.00x      |

**~5.6x faster type checking, ~5.0x faster total compile.** Full unit suite (793 tests) went from
~100 s to 23.8 s; all 793 tests + 22 contest suites still pass.

### DIFF
--- a/sol-core.cabal
+++ b/sol-core.cabal
@@ -95,6 +95,7 @@ library
       Solcore.Frontend.Syntax.Contract
       Solcore.Frontend.Syntax.Location
       Solcore.Frontend.Syntax.Name
+      Solcore.Frontend.Syntax.Traversal
       Solcore.Frontend.Syntax.NameResolution
       Solcore.Frontend.Syntax.Stmt
       Solcore.Frontend.Syntax.SyntaxTree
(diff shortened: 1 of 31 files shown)


mbenke (approved):
Nice work! One question: it seems that `noDesugarOpt` is now not used in code, is it still needed in tests?

Y-Nak (approved):
LGTM

PR #526, opened 2026-07-23
Support ABI decoding of dynamic arrays of ADTs (calldata, lazy)

axic:
## ABI decoding of dynamic arrays of ADTs (calldata, lazy)

Lazily decode a calldata `T[]` whose element `T` is an algebraic data type,
with the ordinary array surface syntax — `ops[i]` and `ops.length()` — and
support for **dynamic** element types (an ADT whose branch carries a
`memory(bytes)`), not just fixed-width static ones.

### Surface syntax
- **`ops[i]`** — a `calldata(array(t)):RValueIdxAccess` instance routes `arr[i]`
  (via `ridx`) to `abiArrayGet`, decoding element `i` on demand. Read-only (no
  `LValueIdxAccess`), so `ops[i] = …` stays a compile error — calldata is
  immutable.
- **`ops.length()`** — a `calldata(array(t)):Length` instance delegates to
  `abiArrayLength`, so length uses the same `Length` UFCS as storage arrays.
  Because `ops` is a parameter, this needed **value-receiver UFCS**: name
  resolution now rewrites `recv.method(args)` → `Class.method(recv, args)` when
  the receiver is a local/parameter and a unique class exposes the method
  (previously only contract-field receivers were rewritten).

### Wire format / decoding
- **Static element** → elements inline, stride `headSize(t)`.
- **Dynamic element** → standard offset table: length word, then one 32-byte
  offset per element (relative to the element region), each pointing at that
  element's own encoding. `abiArrayGet` reads offset `i`, **rebases** a decoder
  onto the element start, and decodes at head-offset 0 — so the element's inner
  offsets (a `memory(bytes)` leaf) resolve relative to the element, which is how
  they were encoded. The existing inline sum/product/`memory(bytes)` decoders
  compose underneath.
- `sum(f,g)`'s ABI **head footprint** is now dynamic-aware: a dynamic sum
  occupies one offset word (32), only a fully static sum is inline `tag + widest
  branch`.
- `abiArrayGet` bounds-checks `i < length`, reverting `ArrayOutOfBounds()`
  (`0x7f52b2bf`).

### Tests
- `abi_array_sum` — dynamic array of a static sum ADT (`Operation`), exercising
  `ops[i]` + `ops.length()`; `tagOf` uses non-trivial sentinels (16/32) so it
  can't pass by echoing the raw wire tag.
- `abi_batch_adt` — nested three-level ADT (`sum`/product, with a `memory(bytes)`
  in one branch → dynamic), exercising the dynamic-element offset-table path.
  Compile test + runtime fixture (`queueSigner`/`execPayload`) in
  `run_contests.sh`.

### Notes
- Inner offsets are **element-relative** (one rebase per array element), a small
  simplification vs strict per-dynamic-value nesting; the two coincide for a
  single dynamic leaf per element.
- Decode-only (returns are `address`/`bytes`, already encodable); dynamic-sum
  *encode* is not implemented here.


std/std.solc
```diff
@@ -1545,6 +1555,50 @@ forall baseType baseType_decoded . ABIDecoder(baseType, CalldataWordReader):ABID
...
+}
+
+// Decode element `i` of a calldata array on demand. Elements sit inline after
+// the length word, each occupying `headSize` bytes, so element `i` starts at
+// (handle + 32) + i * headSize. A fresh element decoder is aimed at the first
+// element and the per-element offset is threaded through as the head offset.
+forall t t_decoded .
+    t : ABIAttribs,
+    ABIDecoder(t, CalldataWordReader):ABIDecode(t_decoded) =>
+function abiArrayGet(a : calldata(array(t)), i : uint256) -> t_decoded {
```
mbenke, line 1610: It seems that `abiArrayGet(ops, i)` with `i >= length` reads past the array into whatever calldata follows (or zeros beyond calldatasize), silently returning garbage instead of reverting the way Solidity's ABI decoder does. Likewise abiArrayLength trusts the length word without checking it fits in remaining calldata.

While `calldata(DynArray)` decode does no validation either,we may leave it be for now, but I think it merits at least a comment.

std/std.solc
```diff
@@ -1545,6 +1555,50 @@ forall baseType baseType_decoded . ABIDecoder(baseType, CalldataWordReader):ABID
...
+     }
+ }
+
+// Length of a decoded calldata array: the handle points at the length word.
+forall t . function abiArrayLength(a : calldata(array(t))) -> uint256 {
+    let rdr : CalldataWordReader = CalldataWordReader(Typedef.rep(a));
+    return uint256(WordReader.read(rdr));
+}
+
+// Decode element `i` of a calldata array on demand. Elements sit inline after
```
mbenke: That's for static elements only, right? What about fynamic ones?

axic: It seems it implemented the fixed-length arrays, the dynamic ones start with length as the first slot. Will fix. This PR focuses on dynamic arrays.

std/std.solc
axic: Added bounds check.

axic:
Here's an example of the ABI encoding of the tail (the head is only a pointer to the tail, as with dynamic types):
```
off     word                       meaning
────────────────────────────────────────────────────────────────────
0x000   0x0000…0002                arrayLength = 2
                                    ── element region starts here (0x020) ──
0x020   0x0000…0040                off₀ = 0x40   (elem₀, relative to 0x020)
0x040   0x0000…0100                off₁ = 0x100  (elem₁, relative to 0x020)

        ── elem₀ = Queue(...)  at 0x020+0x40 = 0x060 ──  (static, inline, 192B)
0x060   0x0000…0000                Batch tag   = 0  (Queue  = inl)
0x080   0x0000…0000                Operation tag = 0 (AddSigner = inl)
0x0a0   A                          address A
0x0c0   0x0000…0001                Signature tag = 1 (Contract = inr)
0x0e0   C                          address C
0x100   0x0000…0000                slack (Signature reserves max-branch width)

        ── elem₁ = Execute(...) at 0x020+0x100 = 0x120 ──  (dynamic)
0x120   0x0000…0001                Batch tag   = 1  (Execute = inr)
0x140   0x0000…0007                n = 7                       (uint256, inline)
0x160   0x0000…0060                bytesOff = 0x60  (relative to elem₁ start 0x120)
0x180   0x0000…0002                bytes length = 2
0x1a0   0xbeef00…00                bytes data (0xbeef, right-padded)
```

mbenke (approved):

PR #528, opened 2026-07-23
Unify a call's result with its expected type

### LINKED ISSUE #527: expected type ignored in some cases
```

forall a.
class a:W { function fromWord(x:word) -> a; }

// presence/absence of this instance seems irrelevant for the error
/*
instance memory(word): W {
  function fromWord(x:word) -> memory(word) { memory(0); }
}
*/
function g(cond: bool, w: word) -> memory(word) {
    if (cond) { return W.fromWord(w); }
    return memory(0);
}
```

fails with

```
error[SC0299]: Ambiguous infered type
  ──> /home/ben/work/solcore/tmp/recquirk.solc:12:10
   │
12 │ function g(cond: bool, w: word) -> memory(word) {
   │          ^ diagnostic reported here
note: forall $2 . $2 : W => bool -> word -> memory(word)
note: in
note: function g (cond : bool, w : word) -> memory(word)
```

Expected behaviour: in `return e`, the type of `e` is unified with the function return type (in this case: `memory(word)`

mbenke:
This PR attempts to fix #527

A class method resolved by its result type was rejected as ambiguous when called from a `return` that is not the last statement of a function:

    function pick(cond: bool, w: word) -> Box(word) {
      if (cond) { return FromWord.fromWord(w); }   // ambiguous a:FromWord
      return Box(w);
    }

Two behaviours combined to produce this.  `tcExpWithExpected'` discarded the expected type for `Call`, unlike `Con` which unifies with it, so the call's result stayed a fresh variable carrying the unresolved constraint.  Nothing else determined it either: `tcBodyWithExpectedReturn` returns only the *last* statement's type upwards, and that is what normally reaches the function level to be unified with the declared return type.  In tail position the second mechanism happened to cover for the first, which is why the same call compiles as the final statement.

`tcCall` now takes the expected result type and unifies with it, mirroring the `Con` case.  This resolves the constraint at the call, independently of the statement's position.  Programs that were wrongly rejected as ambiguous now either compile or report the real error (a missing instance); a non-tail return with a genuinely wrong type is still rejected, as before, since the `Con` path already covered that.

### DIFF
--- a/src/Solcore/Frontend/TypeInference/TcStmt.hs
+++ b/src/Solcore/Frontend/TypeInference/TcStmt.hs
@@ -419,8 +419,8 @@ tcExpWithExpected' _ (FieldAccess (Just e) n) =
     s <- askField tn n
     (ps' :=> t') <- freshInst s
     withCurrentSubst (FieldAccess (Just e') (Id n t'), ps ++ ps', t')
-tcExpWithExpected' _ ex@(Call me n args) =
-  tcCall me n args `wrapError` ex
+tcExpWithExpected' mExpected ex@(Call me n args) =
+  tcCall mExpected me n args `wrapError` ex
 tcExpWithExpected' mExpected (Lam args bd _) =
   do
     (args', schs, ts') <- tcArgs args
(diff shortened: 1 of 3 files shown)


rodrigogribeiro (approved):
Good catch!

PR #529, opened 2026-07-24
Extract array length into a generic Length class

axic:
`length` lived on the `Array` typeclass alongside the storage-only mutators `setLength`/`pop`, which tied the `.length()` UFCS sugar to storage arrays. Split it into its own single-method class `Length` so read-only consumers can depend on just `self:Length`, and so non-storage array representations (e.g. calldata(array(t)) in the ABI layer) can provide `.length()` by implementing one method.

- Add `forall self . class self:Length { length }`; drop `length` from `Array`.
- Move the storage implementation into `instance storage(array(t)):Length`; the `Array` instance keeps `setLength`/`pop`.
- Export `Length` from std.
- Update explicit `Array.length(...)` call sites (and comments) to `Length.length(...)` across the array tests.

UFCS `arr.length()` still resolves uniformly: `length` is now owned by exactly one class, so NameResolution's findClassWithMethod rewrites it to `Length.length(arr)` regardless of where the array lives. The no-context single-parameter instance needs no coverage/Patterson pragmas. Behaviour is unchanged, so the dispatch JSON expectations are untouched.

### DIFF
--- a/test/examples/dispatch/ufcs_array.solc
+++ b/test/examples/dispatch/ufcs_array.solc
@@ -11,7 +11,7 @@ import std.opcodes.{mload, mstore};
 // call is rewritten to Class.method(recv, args).  So:
 //
 //   members.push(addr)  ==>  ArrayPush.push(members, addr)
-//   members.length()    ==>  Array.length(members)
+//   members.length()    ==>  Length.length(members)
 //   members.pop()       ==>  Array.pop(members)
 //
 // It runs against the SAME assertions as storage_array.json (see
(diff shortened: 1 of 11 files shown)


test/examples/dispatch/ufcs_array.solc
```diff
@@ -11,7 +11,7 @@ import std.opcodes.{mload, mstore};
 // call is rewritten to Class.method(recv, args).  So:
 //
 //   members.push(addr)  ==>  ArrayPush.push(members, addr)
-//   members.length()    ==>  Array.length(members)
+//   members.length()    ==>  Length.length(members)
```
axic, line 14: This file tests the UFCS call, the others use the direct class.

mbenke (approved):
LGTM

PR #535, opened 2026-07-31
implementing user defined operators

rodrigogribeiro:


std/std.solc
```diff
@@ -3,6 +3,34 @@ import std.opcodes.{add, sub, mul, div, mod, addmod as addmod_, mulmod as mulmod
...
+infixl 50 (/)  => Div.div;
+infixl 50 (%)  => Mod.mod;
+
+infixl 45 (+)  => Add.add;
+infixl 45 (-)  => Sub.sub;
+
+infixl 44 (&)  => BitAnd.band;
+infixl 43 (^)  => BitXor.bxor;
+infixl 42 (|)  => BitOr.bor;
+
```
axic, line 23: Add bnot please.

test/operators/basic.solc
```diff
@@ -0,0 +1,19 @@
+// Declares a user-defined infix operator (^^) for exponentiation and uses it.
+import std.{*};
+pragma no-patterson-condition ;
+pragma no-coverage-condition ;
+pragma no-bounded-variable-condition ;
+
+infixl 70 (^^) => pow;
```
axic: Solidity uses `**` for this.

std/std.solc
rodrigogribeiro: Added.

axic:
So this could implement `1 ether`, where `ether` is a postfix operator (like `++`) and internally would do `1 * 1 ether`

rodrigogribeiro:
> So this could implement `1 ether`, where `ether` is a postfix operator (like `++`) and internally would do `1 * 1 ether`

Could you give an example of using this operator?

axic:
They are units in Solidity: https://docs.soliditylang.org/en/v0.8.36/units-and-global-variables.html#ether-units

So `1 ether` == `1 * 1e18`.

rodrigogribeiro:
> They are units in Solidity: https://docs.soliditylang.org/en/v0.8.36/units-and-global-variables.html#ether-units
> 
> So `1 ether` == `1 * 1e18`.

Thanks! It is now implemented.

axic:
Awesome, so flexible! These units in classic are a painful exception in the parser 😅

test/operators/basic.solc
rodrigogribeiro: Fixed in the test.

std/std.solc
```diff
@@ -3,6 +3,50 @@ import std.opcodes.{add, sub, mul, div, mod, addmod as addmod_, mulmod as mulmod
 pragma no-patterson-condition ABIEncode, Num, Array, ArrayPush, Eq, Ord;
 pragma no-coverage-condition ABIDecode, MemoryType, Array, ArrayPush, RValueIdxAccess;
 
+// Built-in operators are defined here, not in the compiler. Each declaration
```
axic, line 6: Since these are here -- can we remove the extra parsing stuff from Haskell?

rodrigogribeiro: I believe that I removed these extra parsing stuff from the parsing modules.

axic: Yes, just looked at it properly.

Do wonder what happens in these cases:
1. An operator is not defined, but we try a compound assignment for it.
2. An operator is redefined.
3. An operator is defined as infix and postfix, and other combinations.

Can we have tests for them?

test/operators/redefined-fail.solc
```diff
@@ -0,0 +1,15 @@
+// Redefining an operator symbol is a compile error: `##` is declared twice.
+import std.{*};
+pragma no-patterson-condition ;
+pragma no-coverage-condition ;
+pragma no-bounded-variable-condition ;
+
+infixl 50 (##) => binA;
+infixl 60 (##) => binB;
```
axic, line 8: I'd also add a test case for redefining at the same priority, to make sure none of the parsers make a lookup separation based on that.

src/Solcore/Frontend/Lexer/SolcoreLexer.hs
```diff
@@ -69,9 +71,34 @@ reservedWords =
...
+    "infixr",
+    "infix",
+    "prefix",
+    "postfix"
   ]
 
+isOpChar :: Char -> Bool
+isOpChar c =
+  c `elem` ("+-*/%<>=!&|^~#?" :: String)
+    || (c >= '\x2200' && c <= '\x23FF')
```
axic, line 85: Since these special unicode code pages are allowed, there should be a test using it. Like `1 ∈ arr` for finding a element `1` 😅 

I'm also a bit worried it may be misused, but lets see.

test/operators/redefined-fail.solc
rodrigogribeiro: Added test cases for this.

src/Solcore/Frontend/Lexer/SolcoreLexer.hs
rodrigogribeiro: Well, I could remove that. But, I think it improve readability if it is not overused.

axic:
@rodrigogribeiro opus said this, not sure if any of them are actual bugs:

I'll start by exploring the solcore repo to find the user-defined-operators code.

I couldn't build (no `ghc`/`cabal`/`nix` in this container), so these are from reading the diff `e136159..cd44a4e` (the four UDO commits) — the new `OperatorScan.hs`, the rewritten `Expr.hs` operator table, the `Loader.hs` operator gathering, and the `Stmt.hs` compound-assign desugaring.

## Bugs found

### 1. The operator pre-scan reads comments and string literals — `OperatorScan.hs:31`

```haskell
ops <- many (try opDeclP <|> (anySingle *> pure Nothing))
```

`anySingle` advances **one character** and never skips a comment as a unit (`sc` only runs at the very start and inside `lexeme`). So the scan eventually lands exactly on `infixl` inside a `//` or `/* */` comment, or inside a `"…"` literal, and `keyword "infixl"` matches. The declaration becomes live.

Worst form: because `duplicateOperator` (`OperatorScan.hs:78-84`) compares symbols only — not whole declarations — merely *documenting* an operator breaks the file:

```solc
// The exponentiation operator:
//   infixl 70 (^^) => pow;
infixl 70 (^^) => pow;      // → SC0122: operator (^^) is declared more than once
```

A commented-out `prefix 90 (-) => negate;` in a module that imports `std` instead trips SC0123 via `crossOperatorConflict`. `let s = "infixl 1 (+) => bogus";` silently rebinds `+`. Same defect in `scanImports` (`:114`) and `scanExportModulePaths` (`:208`).

### 2. Import selectors are ignored for operators — `Loader.hs:217-218`

```haskell
selectImportedOperators :: Import -> [OperatorDecl] -> [OperatorDecl]
selectImportedOperators _ ops = ops
```

`gatherImportedOperators` text-scans each directly imported file and takes **every** operator it declares. So `SelectOperator` (parsed at `Decl.hs:132`, then dropped at `NameResolution.hs`, `Loader.hs:687`) does nothing: `import oplib.{pow};` imports `(^^)` just as `import oplib.{pow, (^^)};` does, `hiding {…}` can't exclude an operator, and `import M as N;` brings M's operators in unqualified. Symmetrically, `ExportOperator` is discarded (`expandExportSpecFixed`), so `std.solc`'s `export { (+), (-), … }` list is decorative — and an operator can't be re-exported at all: only *direct* importers see it, never a module that re-exports.

### 3. Operator target names resolve in the *use* site's scope, not the declaring module's

`userRow` (`Expr.hs:72-85`) emits `ExpName Nothing fun [l, r]` with `fun` taken verbatim from the declaration. Nothing ties it to the declaring module, so importing an operator without also importing its bound function under the same name gives an unresolved-name error at the use site:

```solc
import oplib.{(^^)};   // pow not imported
… 3 ^^ 4               // → unknown name `pow`
```

`import std.{*} hiding {and};` likewise breaks `&&` (→ bare `and`). This is why the feature commit had to rewrite `import std;` to `import std.{*};` in `test/examples/comptime/ct_*.solc` — a plain module import no longer makes `+` usable.

### 4. Scanner truncates qualified operator targets to two segments — `OperatorScan.hs:46-53`

```haskell
qualFunP = do
  h  <- identifier
  mt <- optional (char '.' *> identifier)
```

The real parser's `qualifiedName` (`SolcoreTypes.hs:24-27`) takes unboundedly many segments. The operator **table** is built from the scan, so `infixl 45 (<+>) => math.vec.add;` scans clean (the trailing `.add` just falls through to `optional semicolon`) and binds `<+>` to `math.vec`. Silent wrong binding, no diagnostic.

### 5. No check that one precedence level agrees on associativity — `Expr.hs:62-67`

`mergedOpTable` groups purely by numeric precedence. `makeExprParser` tries `pInfixR` then `pInfixL` per level, so a level mixing `InfixL` and `InfixR` stops mid-expression and leaves the rest unconsumed. The repo already sets this up: `test/operators/infixr.solc` declares `infixr 45 (~>)` while `std.solc:17-18` has `infixl 45 (+)` / `(-)`. The test only writes `10 ~> 3 ~> 2`, but `10 ~> 3 + 2` is an unexplainable parse error. Haskell rejects this at declaration time; here it's accepted.

### 6. Match-arm guard uses an empty operator table — `Expr.hs:104`

```haskell
| sym == "|" = notFollowedBy (try (patListP [] *> symbol "=>"))
```

`patListP`'s only use of the operator list is `comptimePatP`, so with `[]` a pattern like `comptime 2 + 3` fails the lookahead, `notFollowedBy` succeeds, and `|` is consumed as bitwise-or instead of the arm separator — misparsing any `match` whose previous arm body ends in a semicolon-less expression. Should be `patListP ops` (thread `ops` through `opSymP`).

### 7. Maximal-munch guard rejects previously valid adjacent operators — `Expr.hs:100-102`

`munchGuard = notFollowedBy (satisfy isOpChar)` for symbolic operators, and `isOpChar` includes `!` and `~`. The old table used `symbol "&&"` with no trailing guard, so `a&&!b`, `x||!y`, `a&~b` all parsed before and no longer do. Nothing in-tree hits it today (I grepped `std/` and `test/`), so it's a latent regression rather than a live break — but the guard needs to exclude symbols that can't start a longer *declared* operator rather than all operator characters.

### 8. `parenOpP` accepts reserved words as operator symbols — `SolcoreLexer.hs:100`

```haskell
identSym = (:) <$> letterChar <*> many identChar
```

No reserved-word check (unlike `identifier`, `:102-108`). `postfix 100 (if) => f;` or `(let)` is accepted and then wrecks statement parsing wherever the keyword appears after an expression.

## Smaller items

- `compoundAssign` / `compoundAssignUnary` (`Stmt.hs:112-127`) match on symbol only, ignoring fixity/arity — a `postfix` `(%)` makes `x %= y` emit a two-argument call to a unary function. They also build `ExpName` without `locatedExpFrom`, so compound assignments lose source spans that infix uses keep.
- `gatherImportedOperators` (`Loader.hs:193-202`) swallows every resolution error with `catchError const (pure [])`, so a mistyped import silently yields no operators and then a wall of unknown-operator parse errors instead of the import diagnostic.
- `scanExportModulePaths` (`OperatorScan.hs:200`) is exported but never called — `Loader.hs:155` uses the parsed `exportModulePaths`. Dead code.
- `std.solc` exports `(!)`…`(||)` but not `(~)` or the eight unit suffixes. Harmless today only because of bug 2; it becomes a real break the moment export filtering is implemented.
- `OperatorScan.hs` duplicates `modPathP` / `externalPathP` / `classifyModulePath` / `splitQual` / `mkQualName` / `itemEntryP` from `Decl.hs:100-135` verbatim. Any future divergence makes the loader resolve a different module for operator gathering than for the real import.

mbenke (commented):
Operator imports are not working correctly, making it difficult to redefine operators:

```
ben@trawa:~/work/review$ cat tmp/udop.solc
import std.{addWord};

infixl 45 (+)  => addWord;

contract Ops {
  function main() -> word { 2 + 2 }
}ben@trawa:~/work/review$ esolc -f tmp/udop.solc -s -g
error[SC0101]: undefined name: Add.add
  ──> /home/ben/work/review/tmp/udop.solc:6:29
  │
6 │   function main() -> word { 2 + 2 }
  │                             ^^^^^^ unknown name
note: in: return Add.add(2, 2) ;
note: in: function main () -> word {
      return Add.add(2, 2) ;
      }
note: in: contract Ops {
      function main () -> word {
      return Add.add(2, 2) ;
      }
      }
note: module validation failed for /home/ben/work/review/tmp/udop.solc
```

rodrigogribeiro:
> @rodrigogribeiro opus said this, not sure if any of them are actual bugs:
> 
> I'll start by exploring the solcore repo to find the user-defined-operators code.
> 
> I couldn't build (no `ghc`/`cabal`/`nix` in this container), so these are from reading the diff `e136159..cd44a4e` (the four UDO commits) — the new `OperatorScan.hs`, the rewritten `Expr.hs` operator table, the `Loader.hs` operator gathering, and the `Stmt.hs` compound-assign desugaring.
> ## Bugs found
> ### 1. The operator pre-scan reads comments and string literals — `OperatorScan.hs:31`
> 
> ```haskell
> ops <- many (try opDeclP <|> (anySingle *> pure Nothing))
> ```
> 
> `anySingle` advances **one character** and never skips a comment as a unit (`sc` only runs at the very start and inside `lexeme`). So the scan eventually lands exactly on `infixl` inside a `//` or `/* */` comment, or inside a `"…"` literal, and `keyword "infixl"` matches. The declaration becomes live.
> 
> Worst form: because `duplicateOperator` (`OperatorScan.hs:78-84`) compares symbols only — not whole declarations — merely _documenting_ an operator breaks the file:
> 
> ```
> // The exponentiation operator:
> //   infixl 70 (^^) => pow;
> infixl 70 (^^) => pow;      // → SC0122: operator (^^) is declared more than once
> ```
> 
> A commented-out `prefix 90 (-) => negate;` in a module that imports `std` instead trips SC0123 via `crossOperatorConflict`. `let s = "infixl 1 (+) => bogus";` silently rebinds `+`. Same defect in `scanImports` (`:114`) and `scanExportModulePaths` (`:208`).
> ### 2. Import selectors are ignored for operators — `Loader.hs:217-218`
> 
> ```haskell
> selectImportedOperators :: Import -> [OperatorDecl] -> [OperatorDecl]
> selectImportedOperators _ ops = ops
> ```
> 
> `gatherImportedOperators` text-scans each directly imported file and takes **every** operator it declares. So `SelectOperator` (parsed at `Decl.hs:132`, then dropped at `NameResolution.hs`, `Loader.hs:687`) does nothing: `import oplib.{pow};` imports `(^^)` just as `import oplib.{pow, (^^)};` does, `hiding {…}` can't exclude an operator, and `import M as N;` brings M's operators in unqualified. Symmetrically, `ExportOperator` is discarded (`expandExportSpecFixed`), so `std.solc`'s `export { (+), (-), … }` list is decorative — and an operator can't be re-exported at all: only _direct_ importers see it, never a module that re-exports.
> ### 3. Operator target names resolve in the _use_ site's scope, not the declaring module's
> 
> `userRow` (`Expr.hs:72-85`) emits `ExpName Nothing fun [l, r]` with `fun` taken verbatim from the declaration. Nothing ties it to the declaring module, so importing an operator without also importing its bound function under the same name gives an unresolved-name error at the use site:
> 
> ```
> import oplib.{(^^)};   // pow not imported
> … 3 ^^ 4               // → unknown name `pow`
> ```
> 
> `import std.{*} hiding {and};` likewise breaks `&&` (→ bare `and`). This is why the feature commit had to rewrite `import std;` to `import std.{*};` in `test/examples/comptime/ct_*.solc` — a plain module import no longer makes `+` usable.
> ### 4. Scanner truncates qualified operator targets to two segments — `OperatorScan.hs:46-53`
> 
> ```haskell
> qualFunP = do
>   h  <- identifier
>   mt <- optional (char '.' *> identifier)
> ```
> 
> The real parser's `qualifiedName` (`SolcoreTypes.hs:24-27`) takes unboundedly many segments. The operator **table** is built from the scan, so `infixl 45 (<+>) => math.vec.add;` scans clean (the trailing `.add` just falls through to `optional semicolon`) and binds `<+>` to `math.vec`. Silent wrong binding, no diagnostic.
> ### 5. No check that one precedence level agrees on associativity — `Expr.hs:62-67`
> 
> `mergedOpTable` groups purely by numeric precedence. `makeExprParser` tries `pInfixR` then `pInfixL` per level, so a level mixing `InfixL` and `InfixR` stops mid-expression and leaves the rest unconsumed. The repo already sets this up: `test/operators/infixr.solc` declares `infixr 45 (~>)` while `std.solc:17-18` has `infixl 45 (+)` / `(-)`. The test only writes `10 ~> 3 ~> 2`, but `10 ~> 3 + 2` is an unexplainable parse error. Haskell rejects this at declaration time; here it's accepted.
> ### 6. Match-arm guard uses an empty operator table — `Expr.hs:104`
> 
> ```haskell
> | sym == "|" = notFollowedBy (try (patListP [] *> symbol "=>"))
> ```
> 
> `patListP`'s only use of the operator list is `comptimePatP`, so with `[]` a pattern like `comptime 2 + 3` fails the lookahead, `notFollowedBy` succeeds, and `|` is consumed as bitwise-or instead of the arm separator — misparsing any `match` whose previous arm body ends in a semicolon-less expression. Should be `patListP ops` (thread `ops` through `opSymP`).
> ### 7. Maximal-munch guard rejects previously valid adjacent operators — `Expr.hs:100-102`
> 
> `munchGuard = notFollowedBy (satisfy isOpChar)` for symbolic operators, and `isOpChar` includes `!` and `~`. The old table used `symbol "&&"` with no trailing guard, so `a&&!b`, `x||!y`, `a&~b` all parsed before and no longer do. Nothing in-tree hits it today (I grepped `std/` and `test/`), so it's a latent regression rather than a live break — but the guard needs to exclude symbols that can't start a longer _declared_ operator rather than all operator characters.
> ### 8. `parenOpP` accepts reserved words as operator symbols — `SolcoreLexer.hs:100`
> 
> ```haskell
> identSym = (:) <$> letterChar <*> many identChar
> ```
> 
> No reserved-word check (unlike `identifier`, `:102-108`). `postfix 100 (if) => f;` or `(let)` is accepted and then wrecks statement parsing wherever the keyword appears after an expression.
> ## Smaller items
> 
>     * `compoundAssign` / `compoundAssignUnary` (`Stmt.hs:112-127`) match on symbol only, ignoring fixity/arity — a `postfix` `(%)` makes `x %= y` emit a two-argument call to a unary function. They also build `ExpName` without `locatedExpFrom`, so compound assignments lose source spans that infix uses keep.
> 
>     * `gatherImportedOperators` (`Loader.hs:193-202`) swallows every resolution error with `catchError const (pure [])`, so a mistyped import silently yields no operators and then a wall of unknown-operator parse errors instead of the import diagnostic.
> 
>     * `scanExportModulePaths` (`OperatorScan.hs:200`) is exported but never called — `Loader.hs:155` uses the parsed `exportModulePaths`. Dead code.
> 
>     * `std.solc` exports `(!)`…`(||)` but not `(~)` or the eight unit suffixes. Harmless today only because of bug 2; it becomes a real break the moment export filtering is implemented.
> 
>     * `OperatorScan.hs` duplicates `modPathP` / `externalPathP` / `classifyModulePath` / `splitQual` / `mkQualName` / `itemEntryP` from `Decl.hs:100-135` verbatim. Any future divergence makes the loader resolve a different module for operator gathering than for the real import.

Thanks! I've fixed all these points (some were really tricky, due to the interaction with modules).

src/Solcore/Frontend/Lexer/SolcoreLexer.hs
axic: No I meant to at least add a test for it and the have it documented later. We can always revise.

rodrigogribeiro: Added tests for unicode operators. After rebasing the new syntax, I'll write about it in the language reference.

rodrigogribeiro:
> Operator imports are not working correctly, making it difficult to redefine operators:
> 
> ```
> ben@trawa:~/work/review$ cat tmp/udop.solc
> import std.{addWord};
> 
> infixl 45 (+)  => addWord;
> 
> contract Ops {
>   function main() -> word { 2 + 2 }
> }ben@trawa:~/work/review$ esolc -f tmp/udop.solc -s -g
> error[SC0101]: undefined name: Add.add
>   ──> /home/ben/work/review/tmp/udop.solc:6:29
>   │
> 6 │   function main() -> word { 2 + 2 }
>   │                             ^^^^^^ unknown name
> note: in: return Add.add(2, 2) ;
> note: in: function main () -> word {
>       return Add.add(2, 2) ;
>       }
> note: in: contract Ops {
>       function main () -> word {
>       return Add.add(2, 2) ;
>       }
>       }
> note: module validation failed for /home/ben/work/review/tmp/udop.solc
> ```

Fixed.

axic:
This is all great, but one question did occur to me (and its not a blocker): how would actual user operators be defined if it is overloading?

i.e. a user wants to define basic arithmetic operations over a specific type (like fixed point numbers), if it is the same precedence (should be) but different function instance. Only through implementing an instance for the classes, and never actually defining a new operator?

rodrigogribeiro:
> This is all great, but one question did occur to me (and its not a blocker): how would actual user operators be defined if it is overloading?
> 
> i.e. a user wants to define basic arithmetic operations over a specific type (like fixed point numbers), if it is the same precedence (should be) but different function instance. Only through implementing an instance for the classes, and never actually defining a new operator?

If it is overloaded, the function most general type is defined in a type class. So, the operator can be defined by referencing to the type class type. As an example, let's consider a type class for addition:

```
forall a . class a : Add {
    function add (x : a, y : a) -> a;
}

infixl 60 (+) => Add.add;
```

The `+` operator will work on any type that has an instance for class Add.

axic:
Yeah that's what I said as one practical case, but for example `!` resolves to a function and not a class instance: `prefix 90 (!)  => not;`

Same with the logical binary operators. Perhaps it is fine, but probably the logical unary operators would be nice having a trait for (can be in a follow up change).

rodrigogribeiro:
> Yeah that's what I said as one practical case, but for example `!` resolves to a function and not a class instance: `prefix 90 (!) => not;`
> 
> Same with the logical binary operators. Perhaps it is fine, but probably the logical unary operators would be nice having a trait for (can be in a follow up change).

Yes, probably the best option is to define a trait / class for each of such operators.

std/std.solc
```diff
@@ -3,6 +3,50 @@ import std.opcodes.{add, sub, mul, div, mod, addmod as addmod_, mulmod as mulmod
...
+infix  35 (<=) => le;
+infix  35 (>=) => ge;
+infix  35 (<)  => lt;
+infix  35 (>)  => Ord.gt;
+
+infix  34 (==) => Eq.eq;
+infix  34 (!=) => ne;
+
+infixl 20 (&&) => and;
+infixl 10 (||) => or;
```
mbenke: These should be short-circuiting, so it might be hard to express as functions

mbenke (changes requested):

std/std.solc
axic: Well, isn't that the "lazy evaluation" discussion again? 😅

rodrigogribeiro: @mbenke is right, we need short-circuiting semantics for this and add lazy evaluation does not seems to be an option.

PR #536, opened 2026-08-01
Add word bitwise NOT with `~` and `~=` sugar

axic:
This follows #485 where this was missed. It uses `~` which is the same as Solidity.

Mirror the binary bitwise operators (`&`/`|`/`^` from #485) for the unary bitwise NOT, reusing the existing `bnotWord` primitive (already constant-folded via the `not` opcode in MastEval):

- std: add a `BitNot` type class with `word` (backed by `bnotWord`) and `uint256` (via `Typedef.rep`/`abs`) instances, and export the class. Method name `bnot` is kept distinct from the boolean `not` used by `!`.
- Parser: `~` prefix operator (same precedence level as `!`, guarded against `~=`) and the unary `~=` compound assignment.
- AST: `ExpBNot` and `StmtBNotEq` (a single operand — it is unary), with name resolution desugaring `~` -> `BitNot.bnot` and `e ~=` to the in-place `e := ~e`.
- Pretty printer and module type-ref renamer handle the new nodes.
- Test: extend cases/bitwise.solc to exercise `~`, `~=`, and the bnotWord constant folding (`~(~x) == x`, `a & ~0 == a`).

`~` (not `!`) matches Solidity, which uses `~` for bitwise NOT and `!` for logical NOT; `!` is already taken here for logical NOT (`ExpLNot`).

### DIFF
--- a/src/Solcore/Frontend/Module/Loader.hs
+++ b/src/Solcore/Frontend/Module/Loader.hs
@@ -1423,6 +1423,8 @@ renameStmtTypeRefs renameMap (StmtBOrEq e1 e2) =
   StmtBOrEq (renameExpTypeRefs renameMap e1) (renameExpTypeRefs renameMap e2)
 renameStmtTypeRefs renameMap (StmtModEq e1 e2) =
   StmtModEq (renameExpTypeRefs renameMap e1) (renameExpTypeRefs renameMap e2)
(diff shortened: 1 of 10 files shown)


rodrigogribeiro (approved):
LGTM

PR #538, opened 2026-08-02
std: make erc7201 compile-time over string literals

axic:
Rewrite `erc7201` to take a string-literal namespace and fold the ERC-7201 storage slot to a `bytes32` constant at compile time, instead of hashing a runtime `memory(bytes)` id. The signature is `-> comptime bytes32`, and the body uses the `-` / `&` / `~` operator sugar now that word bitwise NOT exists:

    function erc7201(id: string) -> comptime bytes32 {
        return bytes32(keccakWordLit(keccakLit(id) - 1) & ~0xff);
    }

The formula keccak256(abi.encode(uint256(keccak256(id)) - 1)) & ~0xff needs a second keccak over the (word-sized) `- 1` preimage, so add a comptime `keccakWordLit` primitive (keccak of a word's 32-byte big-endian form) as the counterpart to the string-literal `keccakLit`. Also give `bnotWord` a folding rule (placed with the other bitwise word ops) so `~0xff` collapses too; both new primitives are registered in MastEval's pure-builtin set.

Tests:
- dispatch/hashes: replace the runtime `erc7201_(bytes)` call with three zero-arg functions returning the comptime slots for "example.main", "openzeppelin.storage.Ownable" (canonical OZ value) and "".
- comptime/erc7201-lit: new folding test for erc7201 + keccakWordLit.

### DIFF
--- a/src/Solcore/Backend/MastEval.hs
+++ b/src/Solcore/Backend/MastEval.hs
@@ -28,7 +28,8 @@ where
 {- Partial Evaluator for Mast
    Performs compile-time evaluation where possible:
    - Interprets assembly blocks containing supported Yul arithmetic operations
-   - Folds calls to `subWord`, `gtWord`, `bxorWord`, `bandWord`, `borWord`, `eqWord` with literal arguments
+   - Folds calls to `subWord`, `gtWord`, `bxorWord`, `bandWord`, `borWord`, `bnotWord`, `eqWord` with literal arguments
+   - Folds the string-literal builtins `concatLit`, `strlenLit`, `keccakLit`, and `keccakWordLit`
    - Propagates known variable values
    - Inlines simple pure functions with literal arguments
 -}
@@ -490,12 +491,15 @@ evalPrimitive (Name "bandWord") [MastLit (IntLit a), MastLit (IntLit b)] =
   Just (MastLit (IntLit (maskWord (a .&. b))))
 evalPrimitive (Name "borWord") [MastLit (IntLit a), MastLit (IntLit b)] =
   Just (MastLit (IntLit (maskWord (a .|. b))))
+evalPrimitive (Name "bnotWord") [MastLit (IntLit a)] =
+  Just (MastLit (IntLit (maskWord (complement a))))
 evalPrimitive (Name "eqWord") [MastLit (IntLit a), MastLit (IntLit b)] =
   Just $ mkBool (a == b)
 -- String literal primitives (Solidity/Yul semantics):
 -- - treat literals as UTF-8 byte sequences
 -- - strlen returns byte length
 -- - keccak returns the 256-bit big-endian word of keccak256(bytes)
+-- - keccakWord return the 256-bit big-endian word of keccak256(be32(word))
 evalPrimitive (Name "concatLit") [MastLit (StrLit a), MastLit (StrLit b)] =
   Just (MastLit (StrLit (a <> b)))
 evalPrimitive (Name "strlenLit") [MastLit (StrLit s)] =
@@ -503,6 +507,13 @@ evalPrimitive (Name "strlenLit") [MastLit (StrLit s)] =
    in Just (MastLit (IntLit (toInteger (BS.length bs))))
 evalPrimitive (Name "keccakLit") [MastLit (StrLit s)] =
   let bs = TE.encodeUtf8 (T.pack s)
+      digest :: Digest Keccak_256
+      digest = hash bs
+      digestBytes :: BS.ByteString
+      digestBytes = BA.convert digest
+   in Just (MastLit (IntLit (bsToIntegerBE digestBytes)))
+evalPrimitive (Name "keccakWordLit") [MastLit (IntLit n)] =
+  let bs = integerToBE32 n
       digest :: Digest Keccak_256
       digest = hash bs
       digestBytes :: BS.ByteString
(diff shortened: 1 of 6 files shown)


src/Solcore/Backend/MastEval.hs
```diff
@@ -490,6 +491,8 @@ evalPrimitive (Name "bandWord") [MastLit (IntLit a), MastLit (IntLit b)] =
   Just (MastLit (IntLit (maskWord (a .&. b))))
 evalPrimitive (Name "borWord") [MastLit (IntLit a), MastLit (IntLit b)] =
   Just (MastLit (IntLit (maskWord (a .|. b))))
+evalPrimitive (Name "bnotWord") [MastLit (IntLit a)] =
```
axic, line 494: Weird that this was missed back when these helpers were introduced.

std/std.solc
```diff
@@ -1088,6 +1089,14 @@ function keccakLit(a:string) -> word {
   return 0;
 }
 
+// keccak256 of a word's 32-byte big-endian representation (comptime-only).
+// The string-literal `keccakLit` hashes UTF-8 bytes; this hashes an already
+// computed word, as needed for the nested keccak of ERC-7201.
+function keccakWordLit(a:word) -> word {
```
axic: Not entirely happy for introducing this, but don't see an easy way around -- basically need a "concat"/"to_bytes" comptime equivalent, therefore I vote for this intermediate solution. 

Maybe a TODO is worth it?

src/Solcore/Backend/MastEval.hs
```diff
@@ -508,6 +511,17 @@ evalPrimitive (Name "keccakLit") [MastLit (StrLit s)] =
...
+-- counterpart of `keccakLit` for an already-computed word (e.g. the outer
+-- keccak of ERC-7201), where the preimage is a 32-byte word rather than a
+-- UTF-8 string literal.
+evalPrimitive (Name "keccakWordLit") [MastLit (IntLit n)] =
+  let bs = integerToBE32 n
+      digest :: Digest Keccak_256
+      digest = hash bs
+      digestBytes :: BS.ByteString
+      digestBytes = BA.convert digest
+   in Just (MastLit (IntLit (bsToIntegerBE digestBytes)))
```
mbenke: avoid repetition, e.g. define

```
keccak256BE :: BS.ByteString -> Integer
keccak256BE bs = bsToIntegerBE (BA.convert (hash bs :: Digest Keccak_256))
```

and then `evalPrimitive` for both keccak ops become one-liners.

std/std.solc
```diff
@@ -1077,6 +1078,14 @@ function keccakLit(a:string) -> word {
...
```
mbenke: why not 


```suggestion
function keccakWordLit(comptime a:word) -> word {
```
this way it should be either folded or rejected at comptime

test/examples/comptime/erc7201-lit.solc
```diff
@@ -0,0 +1,17 @@
...
+  public function main() -> bytes32 {
+    // keccak256(abi.encode(uint256(keccak256("example.main")) - 1)) & ~0xff
+    // == 0x183a6125c38840424c4a85fa12bab2ab606c4b6d0e7cc73c0c06ba5300eab500
+    return erc7201("example.main");
+  }
+
+  // keccakWordLit on its own: keccak of a word's 32-byte big-endian form.
+  public function wordHash() -> word {
+    return keccakWordLit(0);
+  }
```
mbenke, line 16: Perhaps this merits a test: `wordHash` and `keccakWordLit(0)` should return 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563

std/std.solc
```diff
@@ -2702,17 +2711,14 @@ function ecrecover(hash: bytes32, v: uint256, r: bytes32, s: bytes32) -> address
...
-            )
-        ) & ~0xff
-    );
+// ERC-7201 namespaced storage slot, computed entirely at compile time from a
+// string-literal namespace `id`:
+//   keccak256(abi.encode(uint256(keccak256(bytes(id))) - 1)) & ~bytes32(uint256(0xff))
+// `keccakLit` hashes the namespace string and `keccakWordLit` performs the
+// outer keccak over the (word-sized) `- 1` preimage; both keccaks and the word
+// arithmetic fold to a single `bytes32` literal (see MastEval.evalPrimitive).
+function erc7201(id: string) -> comptime bytes32 {
```
mbenke: If it is computed entirely at comptime then I think it should be reflected in the signature:

```
function erc7201(comptime id: string) -> bytes32 {
```

this indicates that `id` has to be known at comptime, otherwise it is a compilation error, and we want that, because otherwise it may slip to runtime and turn into `unimplemented()`. BTW `keccakLit` should most likely have  its param cannotated as comptime as well (I know,this is older code)

OTOH `erc7201(id: string) -> comptime bytes32` indicates that id can be comptime or runtime, but if it is comptime then so is the result.

mbenke (commented):
Nice work, some remarks.

src/Solcore/Backend/MastEval.hs
axic: Done.

std/std.solc

test/examples/comptime/erc7201-lit.solc
axic: Done.

mbenke (approved):

PR #541, opened 2026-08-02
Add EIP-7951 P256VERIFY

axic:
See https://eips.ethereum.org/EIPS/eip-7951

### DIFF
--- a/std/eip7951.solc
+++ b/std/eip7951.solc
@@ -1,0 +1,34 @@
+import std.{*};
+import std.opcodes.{mstore, mload, gas, staticcall};
+
+export { p256verify };
+
+// Perform secp256r1 recovery and check using the P256VERIFY precompile at address
+// 0x100, introduced by EIP-7951 / RIP-7212.
+// Input layout (160 bytes): hash | r | s | qx | qy. The precompile
+// returns a 32-byte word equal to 1 on a valid signature and empty output on an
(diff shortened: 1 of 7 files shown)


test/testrunner/EVMHost.cpp
```diff
@@ -1298,6 +1300,45 @@ evmc::Result EVMHost::precompileBlake2f(evmc_message const&) noexcept
 	return resultWithFailure();
 }
 
+evmc::Result EVMHost::precompileP256Verify(evmc_message const& _message) noexcept
```
axic, line 1303: Lets add this like the others, but I'll work on using evmone-precompiles directly, so these hardcoded values go away for all of them.

std/eip7951.solc
```diff
@@ -0,0 +1,34 @@
+import std.{*};
+import std.opcodes.{mstore, mload, gas, staticcall};
+
+export { p256verify };
+
+// Perform secp256r1 recovery and check using the P256VERIFY precompile at address
```
mbenke, line 6: Does it really perform recovery or just verification?

axic: It does perform recovery internally (as that how ecdsa works), but hides the result apart from the comparison result. We can rephrase.

mbenke (approved):

PR #544, opened 2026-08-03
Add minWord/maxWord

axic:


mbenke (changes requested):
Remove (now duplicated) `maxWord` definiton from `test/examples/cases/abigeneric.solc`

mbenke (approved):

PR #547, opened 2026-08-04
Parse break/continue/leave in inline assembly

axic:
yulStmt's choice had no clause for break/continue/leave, so these Yul
statements were unparseable: `break` inside an asm `for` fell through to
YExp (YIdent "break") and later died with an "undefined name" error.
Every other stage already supported them (tcYulStmt, checkAsmStmt,
renameYulStmts, substYulStmt) and the AST/pretty-printer already had the
YBreak/YContinue/YLeave constructors — only the parser was missing.

Add the three keyword clauses to yulStmt (before the YExp fallback so the
keywords aren't swallowed as identifiers) and a regression test exercising
all three inside an inline-assembly block.

### DIFF
--- a/run_contests.sh
+++ b/run_contests.sh
@@ -7,6 +7,7 @@ cd "$root_dir"
 
 bash ./contest.sh test/examples/dispatch/basic.json
 bash ./contest.sh test/examples/dispatch/assembly.json
+bash ./contest.sh test/examples/dispatch/asm_break_continue_leave.json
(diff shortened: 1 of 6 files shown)


mbenke (approved):
This fails

```
   assembly {
    breakage := add(continued, leaveMeAlone)
    // format := 0
   }
```

but so does `format := 0` so it is preexisting - see #556

PR #550, opened 2026-08-05
Fix some Yul parser bugs

axic:
Fixes three bugs in Language.Yul.Parser, each in its own commit, preceded by a test commit that fails before the fixes and passes after:
- function keyword boundary — yulFun matched with symbol "function" (no word boundary), so functional(x) misparsed as a function definition. Now uses try (pKeyword "function"), letting such identifiers fall through to an expression statement.
- true/false boundary — pKeyword "true"/"false" consumed input before notFollowedBy ran, so trueFlag/falseFlag failed instead of parsing as YIdent. Both are now wrapped in try.
- empty switch — YSwitch used many yulCase, accepting a switch with zero cases; now some yulCase.

New test/YulParserTests.hs covers all three (the switch case as an expected-failure), wired into the sol-core-tests suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sol-core.cabal
+++ b/sol-core.cabal
@@ -198,6 +198,7 @@ test-suite sol-core-tests
     ModuleTypeCheckTests
     SpecialiseTests
     YulEvalTests
+    YulParserTests
     ParserTests
 
   build-depends:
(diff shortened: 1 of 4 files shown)


mbenke (approved):
edit - the previous comment was meant for a different PR

PR #552, opened 2026-08-05
Add `bool:ABIDecode` instance

### LINKED ISSUE #542: std: bool cannot be a public-function parameter (no ABIDecode/ABIAttribs instance for bool)
Affected rev: argotorg/solcore @ 1950e707 (`std/std.solc`).

`bool` has an `ABIEncode` instance (`std/std.solc:1276`), so it works in return position, but there is no `ABIDecode` instance for `bool` (the decode instances at `std/std.solc:1420-1583` cover `uint256`, `bytes32`, `address`, `()`, `memory(string)`, `memory(bytes)`, tuples, and memory/calldata dynamic arrays only) and no `bool` `ABIAttribs` instance. As a result a `bool` cannot appear as a typed public-function parameter.

Repro (fails typecheck at this rev):

```js
import std.{*};
import std.dispatch.{*};
contract C {
    public function f(x : bool) -> () { }
}
```

Removing the parameter, or changing it to `uint256`, typechecks. `public function g() -> bool { return true; }` also typechecks (encode side is fine).

This is already documented as a limitation in `test/examples/dispatch/storage_adt_bool.solc:15-17`, which works around it by taking `uint256` params and converting via a `toBool` helper.

Suggested fix: add an `ABIDecode` instance for `bool` (decode a word and `tobool` it, mirroring the `bool` `ABIEncode` instance) plus a `bool` `ABIAttribs` instance (static, headSize 32), so `bool` is a first-class ABI parameter type in both directions.

axic:
Add an ABIDecoder(bool, reader):ABIDecode(bool) instance so public functions can take bool arguments. bool is a builtin (not a Typedef(word)), so decode round-trips through word via tobool, mirroring the existing bool:ABIEncode instance that uses frombool.

Cover the new instance end-to-end with an id_bool(bool) entry point in the basic dispatch contract and true/false round-trip cases in basic.json.

Fixes #542.

### DIFF
--- a/std/std.solc
+++ b/std/std.solc
@@ -1553,6 +1553,15 @@ forall reader . reader:WordReader => instance ABIDecoder(bytes4, reader):ABIDeco
     }
 }
 
+// ABI Decoding for bool
+// bool is a builtin (not a Typedef(word)), so it round-trips through word via
+// tobool, mirroring the bool:ABIEncode instance which uses frombool.
+forall reader . reader:WordReader => instance ABIDecoder(bool, reader):ABIDecode(bool) {
+    function decode(ptr:ABIDecoder(bool, reader), currentHeadOffset:word) -> bool {
+        return tobool(WordReader.read(WordReader.advance(ptr, currentHeadOffset)));
+    }
+}
+
 // ABI Decoding for address
 forall reader . reader:WordReader => instance ABIDecoder(address, reader):ABIDecode(address) {
     function decode(ptr:ABIDecoder(address, reader), currentHeadOffset:word) -> address {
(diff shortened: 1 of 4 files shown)


std/std.solc
```diff
@@ -1558,7 +1558,10 @@ forall reader . reader:WordReader => instance ABIDecoder(bytes4, reader):ABIDeco
 // tobool, mirroring the bool:ABIEncode instance which uses frombool.
 forall reader . reader:WordReader => instance ABIDecoder(bool, reader):ABIDecode(bool) {
     function decode(ptr:ABIDecoder(bool, reader), currentHeadOffset:word) -> bool {
-        return tobool(WordReader.read(WordReader.advance(ptr, currentHeadOffset)));
+        let v = WordReader.read(WordReader.advance(ptr, currentHeadOffset));
+        // Solidity's ABI decoder rejects a bool whose word is not 0 or 1; match it.
+        require(v <= 1, Error(0x0557dbbf)); // DirtyHigherBitsForBool()
```
axic, line 1562: Solidity also has a strict decoder.

mbenke (approved):

PR #553, opened 2026-08-05
Clear ecrecover return-data scratch before the precompile call

### LINKED ISSUE #543: std: ecrecover can return a bogus signer on invalid recovery (result read from un-zeroed scratch)
Affected rev: argotorg/solcore @ 1950e707 (`std/std.solc`, function `ecrecover` at lines 2551-2573).

`ecrecover` directs the precompile (address 0x01) output to scratch `(0, 32)` and then reads the recovered signer with `mload(0)`:

```js
let ret = staticcall(gas(), 1, ptr, 128, 0, 32);
require(ret != 0, Error(0x578763f7)); // ECRecoverCallFailed()
let res = mload(0);
require(res != 0, Error(0x4fbfae63)); // ECRecoverFailed()
return address(res);
```

The scratch slot `[0,32)` is never pre-zeroed. On an invalid recovery (v not in {27,28}, or an r/s that recovers no point - the malleability guard only screens s), the precompile returns success (`ret != 0`) but writes zero bytes of output, so the EVM leaves `[0,32)` untouched. `mload(0)` then reads whatever was in scratch from a prior operation. When that stale word is nonzero, `require(res != 0)` passes and `ecrecover` returns a garbage address instead of reverting `ECRecoverFailed()`. solc's own ecrecover codegen pre-zeroes this buffer; std does not.

**Reproducer** (minimal, self-contained; verified against this rev). Both functions make the SAME invalid-signature call as std's own `recoverFail()` example (r = 0), differing only in whether scratch `[0,32)` is dirtied first:

```js
import std.{*};
import std.opcodes.{mstore};
import std.dispatch.{*};

contract EcrecoverScratchBug {
    constructor() {}

    // Scratch clean: mload(0) reads 0, ecrecover reverts ECRecoverFailed().
    public function recoverFailClean() -> address {
        return ecrecover(
            bytes32(0xaabbccddeeff00112233445566778899aabbccddeeff00112233445566778899),
            27, bytes32(0x0),
            bytes32(0x3523e7d34da277c59af090e44cebddb10b73be11780f028d02cf5ae5f24109fc));
    }

    // Same invalid signature, scratch [0,32) dirtied first: the empty
    // precompile return leaves the dirty word in place, so ecrecover reads it
    // back and returns it as a bogus signer instead of reverting.
    public function recoverFailDirtyScratch() -> address {
        mstore(0, 0x000000000000000000000000deaddeaddeaddeaddeaddeaddeaddeaddeaddead);
        return ecrecover(
            bytes32(0xaabbccddeeff00112233445566778899aabbccddeeff00112233445566778899),
            27, bytes32(0x0),
            bytes32(0x3523e7d34da277c59af090e44cebddb10b73be11780f028d02cf5ae5f24109fc));
    }
}
```

Observed (compiled with sol-core/yule/solc at this rev, deployed on anvil):

```js
recoverFailClean()         -> reverts: custom error 0x4fbfae63 (ECRecoverFailed)   [correct]
recoverFailDirtyScratch()  -> returns: 0xdeaDDeADDEaDdeaDdEAddEADDEAdDeadDEADDEaD   [BUG]
```

The dirty function returns the exact word planted in scratch as the "signer", confirming the stale-scratch leak. As a dispatch test-example vector, the correct expectation for BOTH is a `4fbfae63` failure; the current std produces that only for `recoverFailClean` and wrongly succeeds for `recoverFailDirtyScratch`:

```json
{ "input": { "comment": "recoverFailClean()(address)",        "calldata": "a07c2d44", "value": "0" },
  "kind": "call", "output": { "returndata": "4fbfae63", "status": "failure" } }
{ "input": { "comment": "recoverFailDirtyScratch()(address)", "calldata": "d00c9bf3", "value": "0" },
  "kind": "call", "output": { "returndata": "4fbfae63", "status": "failure" } }
```

The existing `test/examples/dispatch/ecrecover.solc` `recoverFail()` does NOT catch this: it never writes to `[0,32)` before the call, so scratch is coincidentally zero and the ECRecoverFailed revert fires. Its comment ("`res` stays 0") assumes the buggy invariant. A regression test should dirty scratch first, exactly as `recoverFailDirtyScratch` above.

**Impact:** any contract that does a memory/mapping/keccak operation touching `[0,32)` before an invalid `ecrecover` can silently accept a forged/malformed signature as a valid signer instead of reverting. (Found the issue while doing a side project in Core Solidity over the weekend, which shims with a `mstore(0,0)` then `ecrecover` wrapper; its regression test is mutation-verified to fail against the raw std `ecrecover`.)

**Suggested fix:** add `mstore(0, 0)` immediately before the `staticcall` in `std/std.solc` `ecrecover`, so an empty precompile return reads back as 0 and the `require(res != 0, ECRecoverFailed())` guard fires as intended. Also add a regression test that pre-dirties `[0,32)`.

axic:
The ecrecover precompile reports success (ret != 0) even when it recovers nothing — e.g. when v is not 27 or 28, or in the generic could-not-recover case — but in that case it returns empty output, so the [0, 32] output area passed to staticcall is left untouched. `res = mload(0)` then reads whatever stale word was already there; a non-zero value slips past the `res != 0` guard and ecrecover returns a bogus address instead of reverting.

Add `mstore(0, 0)` before the call so the scratch area reads back as 0 on an empty return, making the `ECRecoverFailed()` guard fire reliably.

Add a `recoverFailBadV()` test (v = 1, well-formed r/s) covering the v not in {27, 28} path; the existing `recoverFail()` (r = 0) test covers the generic could-not-recover path.

Fixes #543.

### DIFF
--- a/std/std.solc
+++ b/std/std.solc
@@ -2565,6 +2565,12 @@ function ecrecover(hash: bytes32, v: uint256, r: bytes32, s: bytes32) -> address
     mstore(ptr + 32, v_);
     mstore(ptr + 64, r_);
     mstore(ptr + 96, s_);
+    // Clear the [0, 32] scratch space that receives the return data. On a
+    // failed recovery (e.g. v not in {27, 28}, or the generic could-not-recover
+    // case) the precompile still reports success but returns no data, leaving
+    // the output area untouched. Without this, a stale non-zero value would
+    // slip past the `res != 0` check below and yield a bogus address.
+    mstore(0, 0);
     let ret = staticcall(gas(), 1, ptr, 128, 0, 32);
     require(ret != 0, Error(0x578763f7)); // ECRecoverCallFailed()
     let res = mload(0);
--- a/test/examples/dispatch/ecrecover.json
+++ b/test/examples/dispatch/ecrecover.json
@@ -34,6 +34,18 @@
           "returndata": "4fbfae63",
           "status": "failure"
         }
+      },
+      {
+        "input": {
+          "comment": "recoverFailBadV()(address)",
+          "calldata": "5dc458b7",
+          "value": "0"
+        },
+        "kind": "call",
+        "output": {
+          "returndata": "4fbfae63",
+          "status": "failure"
+        }
       }
     ]
   }
--- a/test/examples/dispatch/ecrecover.solc
+++ b/test/examples/dispatch/ecrecover.solc
@@ -21,4 +21,19 @@ contract EcrecoverTest {
         let s: bytes32 = bytes32(0x3523e7d34da277c59af090e44cebddb10b73be11780f028d02cf5ae5f24109fc);
         return ecrecover(h, v, r, s);
     }
+
+    // v = 1 is not a valid recovery id (only 27 and 28 are accepted). The
+    // precompile still succeeds (ret != 0) but returns empty output, leaving the
+    // [0, 32] return-data area untouched. Only because `ecrecover` now clears
+    // that area with `mstore(0, 0)` before the call does `res` reliably read as
+    // 0 and hit the `ECRecoverFailed()` (0x4fbfae63) revert path — without the
+    // clear a stale non-zero word would be returned as a bogus address. `r` and
+    // `s` are the well-formed values from `recover()` so only `v` is at fault.
+    public function recoverFailBadV() -> address {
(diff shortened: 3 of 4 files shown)


rodrigogribeiro (approved):
LGTM

PR #554, opened 2026-08-05
Isolate contest runner artifacts

Y-Nak:
Give each contract-test invocation its own output location.

Fixes artifact collisions where local runs could overwrite the same generated Hull files.


mbenke (approved):
Nicely done

PR #559, opened 2026-08-10
Fix parser `pKeyword` backtracking in Yul

### LINKED ISSUE #556: Yul keywords are to greedy
```
ben@trawa:~/work/solcore$ cat tmp/yulkw.solc
function bad() -> () {
   format : word
   assembly {
     format := 0
   }
}ben@trawa:~/work/solcore$ cabal exec sol-core -- -f tmp/yulkw.solc 
error[SC0001]: parse error: unexpected 'm'
  ──> /home/ben/work/solcore/tmp/yulkw.solc:4:9
  │
4 │      format := 0
  │         ^ unexpected token
```

Proposed fix: 

pKeyword is not backtracking:

``` haskell
pKeyword w = lexeme (string w <* notFollowedBy identChar)
```

change to

``` haskell
pKeyword w = lexeme (try (string w <* notFollowedBy identChar))
```

axic:
Fixes #556.

The Yul parser's pKeyword consumed input on a partial match, so the
statement-parser choice (let/if/for/switch/break/continue/leave, plus
default/case) could fail to fall through to a later alternative. Wrap it
in try, matching the already-correct Hull/Parser.hs.

### DIFF
--- a/test/YulParserTests.hs
+++ b/test/YulParserTests.hs
@@ -31,7 +31,8 @@ yulParserTests =
     "Yul parser"
     [ functionKeywordTests,
       boolLiteralTests,
-      switchTests
+      switchTests,
+      statementKeywordTests
     ]
 
 -- Bug 1: yulFun matched the `function` keyword with `symbol "function"`, which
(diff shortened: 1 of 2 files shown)


test/YulParserTests.hs
```diff
@@ -70,6 +71,40 @@ boolLiteralTests =
...
+-- with a keyword (e.g. `format` vs `for`) failed to parse. Every
+-- statement-level keyword is affected, including the merged
+-- break/continue/leave.
+statementKeywordTests :: TestTree
+statementKeywordTests =
+  testGroup
+    "statement keyword boundary"
+    ( [ testCase (kw ++ "-prefixed identifier parses as a call statement") $
+          parsesAs yulStmt (kw ++ "ish") (YExp (yIdent (kw ++ "ish")))
+      | kw <- ["let", "if", "for", "switch", "break", "continue", "leave"]
```
mbenke: Perhaps add `case` `default` for completeness

axic: @mbenke done

mbenke (approved):

PR #565, opened 2026-08-11
ADT ABI wire format: bytes32 keccak variant tag instead of inl/inr 0/1 chain

axic:
Change the on-wire encoding of a multi-constructor ADT that crosses the ABI
boundary so its discriminant is a single bytes32 tag —
keccak256("Name(argSigs...)") — instead of the positional inl/inr 0/1 chain.
The tag reuses the SigString convention (std.dispatch) for the argument types,
exactly the shape a method selector hashes but kept as a full 32-byte word.
This is ABI-wire-only: the internal Hull/Yul sum representation and storage
encoding are unchanged.

The variant name is gone by the structural sum(f,g) level, so DeriveGeneric
now emits per-type, variant-aware ABIEncode + ABIDecode for multi-constructor
types (the encoder can no longer be the Generic bridge, which would fall back
to the anonymous sum's 0/1 tag). Single-constructor ADTs are products with no
discriminant and keep the plain bridge (no tag added).


rodrigogribeiro (approved):
Great work!

PR #566, opened 2026-08-11
Exclude memory-op asm from the comptime purity notion

axic:
`asmIsInterpretable` treats `mload`/`mstore`/`mstore8` as interpretable so that memory-touching functions enter `pureFuns`. That set feeds two very different consumers:

  1. Partial evaluation inlining, where memory folding is sound because the Yul interpreter only executes memory ops under `envComptimeMode` and `mload` succeeds only for bytes written earlier in the same evaluation context (dynamic gating).

  2. `ComptimeCheck.isComptime`, which classifies expressions *structurally* with no such gating. A call to an `mload`-reading function that partial evaluation could not fold survives to the comptime check and, because it was in `pureFuns`, was wrongly classified comptime — even though its result depends on runtime memory. That let a `let x : comptime word` binding accept a genuinely runtime memory read.

Split purity into two notions selected by a `memOk` flag:

  * `computePureFuns` (memOk = True)  — unchanged, drives PE.
  * `computeComptimePureFuns` (memOk = False) — excludes memory-op functions; now used by `ComptimeCheck`.

A legitimately foldable memory chain (mstore then mload of the same address, `ct_asm_mem.solc`) is already collapsed to a literal by PE before the comptime check runs, so it is unaffected. An unfoldable memory read survives as a call and is now correctly rejected (new negative test `ct_asm_mload_runtime.solc`).

Also generalise `asmIsInterpretable` to `asmIsInterpretableWith memOk`, add strict-variant unit tests, and update `doc/comptime-asm.md`.

### DIFF
--- a/doc/comptime-asm.md
+++ b/doc/comptime-asm.md
@@ -25,15 +25,44 @@ type YulState = Map.Map Name Integer  -- variable bindings only (no memory yet)
 
 ### Purity classification
 
-`asmIsInterpretable :: [YulStmt] -> Bool` is used by `stmtIsPure` to decide whether
-a function whose body contains an asm block belongs in `pureFuns`.  This is load-bearing
-for comptime-check soundness: a function containing `sload` must **not** enter `pureFuns`,
-or `isComptime` in `Backend/ComptimeCheck.hs` would wrongly classify its result as comptime.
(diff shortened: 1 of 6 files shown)


axic:
#563 triggered one of these bugs

mbenke (approved):

PR #567, opened 2026-08-11
Desugar field accesses nested inside constructor arguments

axic:
The field-access desugarer's transRhs left constructor applications
untouched (`go Con{} = pure e`), so a contract field access sitting inside
a constructor's arguments in r-value position was never lowered to its
storage-access form. Tuple literals parse to `pair(..)` constructors, so
`concat((this.branch[i], node))` carried an undesugared `this.branch[i]`
through to type checking, which failed with
"tcExp not implemented for: this.branch".

Recurse into constructor arguments like Call/ArrayLit already do. Nullary
constructors are unaffected (mapM over []), and constructor args in r-value
position should be desugared regardless, so this only fixes previously
broken cases.


rodrigogribeiro (approved):
LGTM.

PR #568, opened 2026-08-11
Add Solidity-style structs with named fields and dot access

axic:
## Add Solidity-style structs (named fields + dot access)

Adds a `struct` type to the frontend — a record with named fields, read via dot notation:

```
struct Point {
  x: uint256;
  y: uint256
}

let p : Point = Point(1, 2);
let a : uint256 = p.x;
```

### Approach

A struct is represented internally as a **single-constructor product** — `struct Point { x: uint256; y: uint256 }` desugars to `data Point = Point(uint256, uint256)` whose constructor also carries the field names. Because a struct is just a *tagless product*, it inherits existing machinery with **no backend, Mast, or Hull changes**:

- **ABI encode/decode** is exactly Solidity's tuple wire format (via the `Generic` / `ABIGeneric` derivation already used for ADTs).
- **Backend lowering** is the existing SoA / `fst`-`snd` product lowering.

Field reads `s.x` lower to generated positional **projection functions**, resolved by the type of `s` (so same-named fields on different structs disambiguate).

### What's in the frontend

- `Constr` (both AST layers) gains `constrFields :: [Name]` (empty for ordinary `data`), carried through name resolution and preserved by the module loader / substitution.
- New `struct` keyword + parser production. Field syntax is postfix `name : type`, matching contract fields and function params; fields are `;`-separated with an optional trailing `;`.
- Name resolution registers struct field labels and resolves `recv.field` on a value receiver to member access; the early contract-field desugarer passes those nodes through.
- New `Desugarer.StructProjection` pass emits one projection per field; `TcStmt` rewrites `e.field` to the matching projection call.
- `checkDataType` records field names so the access type-checks; annotated top-level function signatures are pre-registered before bodies so the type-check-introduced projection calls resolve regardless of SCC order.

### Tests

- **`ParserTests`**: struct declaration parses to the expected single-constructor product.
- **`test/examples/cases/struct_fields.solc`, `struct_abi.solc`**: compile-through-pipeline coverage of construction, single-level and nested field reads (`l.p1.x`), structs as params/returns, and `abi_encode` of a struct.
- **`test/examples/dispatch/abi_struct.{solc,json}`**: testrunner end-to-end test exercising both ABI directions — encode, decode+encode round-trip of a struct parameter, and decode + dot-access. The struct includes a nested ADT field (`color : Option(Color)`) to exercise a struct whose field is itself a (static) sum type.

### Notes for review

- The `abi_struct` dispatch fixture's selectors/returndata were computed from the ADT ABI wire-format rules (the keccak variant-tag model was validated against the existing `abi_encode_adt` fixtures); worth a sanity check against a fresh `run_contests.sh` run.
- Scope is **value/ABI structs**; storage-backed struct l-values (`s.x = v` into a slot) are not included and are a natural follow-up (the `Generic` storage derivation already handles the product side).


axic:
Syntax should be `x: type;` before the big syntax overhaul.

rodrigogribeiro (approved):
Great work.

PR #571, opened 2026-08-17
Support struct sub-field writes to storage fields

axic:
For a struct-typed contract (storage) field:

- Read `f.x`: the field desugars to `RVA.acc(...)`, a type-class method whose
  result type is only fixed by constraint solving, too late for the field
  projection which needs the struct type at the access site. Annotate the loaded
  value with the field's declared type — `(RVA.acc(f) : S).x` — so the receiver
  type is concrete when TcStmt lowers the projection.
- Write `f.x = v`: desugars to a whole-struct read-modify-write `f = set_x(f, v)`.
  StructProjection generates, per field, a setter `set_x(s, v) -> s` that
  reconstructs the struct with that field replaced, alongside the projections
  (exposing fieldSetName and structFieldMap). FieldAccess threads a struct ->
  field-names map through the contract env and rewrites the nested-FieldAccess
  assignment LHS to the whole-field storage write (Assign / CanStore.store). No
  new storage machinery; the setter call is emitted pre-typecheck so SCC orders
  it normally.

Local (non-storage) struct variable field writes remain out of scope.

Note: field-based writes write everything. As a follow up we need to design optimised ways to read/write fields, i.e. only emit partial sstore/sloads.


rodrigogribeiro (approved):
LGTM.

PR #575, opened 2026-09-03
Some changes to the compiler / ERC20 without inheritance using ERC7201

rodrigogribeiro:


czepluch (commented):
shouldn't it be named "vault" instead of "valt"? The directory for the vaulttoken

test/examples/erc20/mytoken.sol
```diff
@@ -0,0 +1,118 @@
...
+  let c : word;
+  assembly { c := chainid() }
+  return uint256(c);
+}
+function nowTime() returns (uint256) {
+  let t : word;
+  assembly { t := timestamp() }
+  return uint256(t);
+}
+
```
czepluch: why not use std.opcodes and get rid of assembly?

test/examples/erc20/erc20core.sol
```diff
@@ -0,0 +1,98 @@
...
+    let ts : uint256 = HasSupply.totalSupply(s);
+    let nts : uint256 = Num.add(ts, value);
+    require(nts >= ts, Error(0x4e487b71));       // overflow -> Panic(0x11)-like
+    HasSupply.setTotalSupply(s, nts);
+  } else {
+    let bf : uint256 = HasLedger.balanceOf(s, from);
+    require(bf >= value, Error(0xe450d38c));     // ERC20InsufficientBalance
+    HasLedger.setBalance(s, from, Num.sub(bf, value));
+  }
+  if (to == address(0)) {
```
czepluch, line 105: `transfer(to = address(0))` reaches this burn branch: nothing on the public `transfer/transferFrom` path rejects a zero recipient, so a typo'd recipient burns the tokens and reduces supply. OZ reverts here with `ERC20InvalidReceiver`. Guard the entry points, or make burn explicit in the seam `(Option<address>)` so it cannot happen by accident. Same issue in `valt/erc20base.sol:47`.

test/examples/erc20/erc20core.sol
```diff
@@ -0,0 +1,98 @@
+// ERC20 core capability, with its own storage (three sub-slots of an ERC-7201 root).
+import * from std;
+import {sload, sstore, log3, mstore} from std.opcodes;
+import * from store;
+
+export { HasLedger, HasSupply, HasAllowance, coreUpdate, spendAllowance, approveVal };
```
czepluch, line 7: These exported setters plus a freely constructible `AppStore` let any code bypass `tokenUpdate` (pause check, votes, `Transfer` event). Consider exporting a single mutator that takes the hook evidence as a parameter. same enforcement OZ gets from `private _balances`.

test/examples/erc20/mytoken.sol
rodrigogribeiro: Fixed.

test/examples/erc20/erc20core.sol
rodrigogribeiro: Fixed.

test/examples/erc20/erc20core.sol
rodrigogribeiro: Fixed.

